Issue #11, opened 2019-07-10
A better name


JaggerJo:
Fable and Fabulous are great sounding names. FuncUI is ok, but not as great.

If you have suggestions please comment below.

kMutagene:
You could go the way of [FAKE (F# MAKE)](https://fake.build/), using the original name Avalonia and prepend an F for F#: Favalonia. Or like [Fable (F# |> Babel)](https://fable.io/), additionally altering the original name a bit, which could lead to something like Favalanche. Those are the first ideas from the top of my mind. 

Additionally, there are always [Acronym generators](https://acronymify.com/), although some suggestions are not the best suited ones: 
Top scoring suggestion: AFIRE | Avalonia Functional InteRfacE
One of the bad ones: ANUS | Avalonia fuNctional USer 😆

JaggerJo:
Have thought about it for a bit, and think the name is good. It describes really god what the library is about after all.

Also if the library gets extended to other frameworks (WPF, UWP, Xamarin) the name still works great. The prefix can changed accordingly.

I really want to like Fabulous but don’t like how events are abstracted away and how you are still using MVVM constructs (you use command instead of click, ...)

PR #12, opened 2019-07-10
Add dotnet template folder and template project for ElmishTemplate

kMutagene:
This PR contains the minimum structure for a dotnet template. I used the elmish project from the examples folder to setup this template. You still would have to install it locally at this point though. The next steps should be improving on this template and then publishing it via nuget so you can directly install it from the web.

Screenshot of me installing the template from the project folder:

image

Creating the project:

image

Running the counter example after successfully building it:
image

### DIFF
--- a/src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
+++ b/src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
@@ -3,16 +3,17 @@
   "author": "Josua Jäger, Kevin Schneider",
   "classifications": [
     "Console",
+    "Avalonia",
     "UI",
     "Elmish"
   ],
-  "name": "Elmish Template for Avalonia.FuncUI v0.1.0",
+  "name": "Avalonia.FuncUI cross-plattform MVU GUI App",
   "tags": {
     "language": "F#",
     "type": "project"
   },
-  "identity": "Avalonia.FuncUI.ElmishTemplate",
-  "shortName": "avaloniaElmish",
-  "sourceName": "Avalonia.FuncUI.ElmishTemplate", // Will replace the string 'Avalonia.FuncUI.ElmishTemplate' when using dontet -n.
+  "identity": "Avalonia.FuncUI.BasicTemplate",
+  "shortName": "avalonia-mvu",
+  "sourceName": "Avalonia.FuncUI.BasicTemplate", // Will replace the string 'Avalonia.FuncUI.Basic' when using dontet -n.
   "preferNameDirectory": true
 }
--- a/src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
+++ b/src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "http://json.schemastore.org/template",
+  "author": "Josua Jäger, Kevin Schneider",
+  "classifications": [
+    "Console",
+    "UI",
+    "Elmish"
+  ],
+  "name": "Elmish Template for Avalonia.FuncUI v0.1.0",
+  "tags": {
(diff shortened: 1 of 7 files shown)


src/templates/Avalonia.FuncUI.ElmishTemplate/content/Counter.fs
```diff
@@ -0,0 +1,60 @@
+﻿namespace CounterElmishSample
+
+open Avalonia.Controls
+open Avalonia.Media
+open Avalonia.FuncUI.Types
+open Avalonia.FuncUI
+open Avalonia.Layout
+
+type CustomControl() =
```
JaggerJo: shouldn't be included in the basic template.

src/templates/Avalonia.FuncUI.ElmishTemplate/content/Counter.fs
```diff
@@ -0,0 +1,60 @@
...
+    inherit Control()
+
+    member val Text: string = "" with get, set
+
+[<AutoOpen>]
+module ViewExt =
+    type Views with
```
JaggerJo: so this can be also removed

src/templates/Avalonia.FuncUI.ElmishTemplate/content/Main.fs
```diff
@@ -0,0 +1,59 @@
...
+
+type MainWindow() =
+    inherit HostWindow()
+
+    do
+        base.Title <- "Counter Elmish"
+        base.Height <- 400.0
+        base.Width <- 400.0    
+      
+        //this.VisualRoot.VisualRoot.Renderer.DrawFps <- true
```
JaggerJo: should this be included ?

src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
```diff
@@ -0,0 +1,18 @@
...
+    "Console",
+    "UI",
+    "Elmish"
+  ],
+  "name": "Elmish Template for Avalonia.FuncUI v0.1.0",
+  "tags": {
+    "language": "F#",
+    "type": "project"
+  },
+  "identity": "Avalonia.FuncUI.ElmishTemplate",
```
JaggerJo: Not sure about the name. 

I think 'Elmish' should be included in every template, so I would ditch the prefix and just call it 'basic'/'app' ?

In the future there could also be an 'Empty' template containing no view code, but has the rest already setup.

What do you think ? 🤔

src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json.schemastore.org/template",
+  "author": "Josua J�ger, Kevin Schneider",
+  "classifications": [
+    "Console",
+    "UI",
+    "Elmish"
+  ],
+  "name": "Elmish Template for Avalonia.FuncUI v0.1.0",
```
JaggerJo: remove the version and update the name (if it is visible) ?

src/templates/Avalonia.FuncUI.ElmishTemplate/content/.template.config/template.json
kMutagene: > I think 'Elmish' should be included in every template, so I would ditch the prefix and just call it 'basic'/'app' ?

Will do. Due to the name Avalonia.FuncUI i have a hard time coming up with something really short for the "shortName" field. Will not be an issue anymore when #11 is closed.

> In the future there could also be an 'Empty' template containing no view code, but has the rest already setup.

Speaking from experience using the safe stack template (which uses MVU with elmish to create webapps) it is very helpful to start with a small working example to get a hold of the MVU cycle. However, a --empty argument for the template should work.

PR #21, opened 2019-10-03
sketch for adaptive funcui

dsyme:
This is micro sketch for how FuncUI v1 might use adaptive data

### DIFF
--- a/src/Avalonia.FuncUI/Avalonia.FuncUI.fsproj
+++ b/src/Avalonia.FuncUI/Avalonia.FuncUI.fsproj
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="Avalonia.Desktop" Version="0.8.0" />
+    <PackageReference Include="FSharp.Data.Adaptive" Version="0.0.2" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


dsyme:
This is not close to complete or even convincing, just putting the basics there to define an `adaptivePatch` function that, when the return closure is invoked, looks for changes in the adaptive data and applies them as a patch 

https://github.com/JaggerJo/Avalonia.FuncUI/pull/21/files#diff-ab3843d5f352bde4083a9cc0f90d3509R565

JaggerJo:
Thanks Don.

Will look into this next week. Would be really cool to make diffing cheaper

dsyme:
> Will look into this next week. Would be really cool to make diffing cheaper

Just to mention that it does come with costs in the programming model for adaptive views. And also there is an underlying issue about whether to use immutable or mutable data for the model.  (The Aardvark people often use an immutable model with a reflection-based technique to diff the model to translate into modifications to ChangeableValue etc., which then cause the incremental view diff'ing to kick in.)

JaggerJo:
Just had the time to look into this a bit further, and really like the concept. Getting the API that is exposed to the user simple would be the most challenging part IMO, but maybe I'm wrong. 

I don't think it's required for most applications, because view diffing with some tricks is fast enough in most cases (as far as my experience goes). If vNext is ready I might try out a hybrid approach, so one could use an adaptive view for that huge chart, but still use the expressive and simple views for the rest.

PR #22, opened 2019-10-04
Designed a new logo for Avalonia FuncUI

0xaryan:
I've designed a new logo for FuncUI which looks similar to Avalonia's logo, but more F#y.

Avalonia Logo:

avalonia

FuncUI Logo:

FuncUI_Boxed

Readme updated as well:

image


JaggerJo:
Wow, good job!

Happypig375:
>That looks F#y

Substitute # = ish
That looks Fishy🐟

PR #32, opened 2019-10-26
Feature/dsl

uxsoft:
ToolTip + UniformGrid


src/Avalonia.FuncUI.DSL/ToolTip.fs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// A value indicating whether the tool tip is visible.
+        /// </summary>
+        static member isOpen<'t when 't :> Control>(value: bool) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<bool>(ToolTip.IsOpenProperty, value, ValueNone)
+       
+        /// <summary>
+        /// The content to be displayed in the control's tooltip.
+        /// </summary>
+        static member tip<'t when 't :> Control>(value: obj) : IAttr<'t> =
```
JaggerJo: ToolTips can contain other controls as a 'tip'. This is similar to ContentControls content property. see https://github.com/JaggerJo/Avalonia.FuncUI/blob/927b5dcd158d7cc9f7c742b04bfd355a9ff89521/src/Avalonia.FuncUI.DSL/ContentControl.fs#L12

uxsoft: I added 3 more overloads as its in the ContentControl for the tip property

PR #53, opened 2020-01-20
Add Textbox Watermark

AngelMunoz:
Hello there!

<details>
  <summary>
  First of all, this is such an awesome project :sunglasses:
  </summary
  <p>
    I've been really enjoying doing some toy apps with this, and next time I'm doing something for desktop I'm sure I'll pick this to work with
  </p>
</details>

I noticed the watermark property was missing from the textbox and I decided to give it a shot, I added also a page in the control catalog project with a small watermark sample, but I'm not sure if that's a good place to put it or should it go as a different example Please let me know where should it go and I can make the changes also if I should add more to it rather than just a textbox with watermarks


JaggerJo:
Hi, thanks for the PR! 

Your changes look good to me, I will publish a new package this week.

Happy to hear that you like it! I've seen your AvaFunc test app the FSharp Weekly newsletter. Let me know if you run into some issues or if you have other feedback.

The sample Control Catalog is a bit of a mess right now, so I thinks it's ok to put the text box example there.

AngelMunoz:
>I've seen your AvaFunc test app the FSharp Weekly newsletter.
That's nice I didn't notice I guess I'll have to polish it a little bit more

PR #56, opened 2020-01-22
Calendar, DatePicker bindings

AngelMunoz:
Hello there,
I gave it another shot at the bindings I tried to come up with the Calendar and DatePicker Bindings
I added a sample to the control catalog
I just have a question about some of the properties

```fsharp
static member selectedDate<'t when 't :> Calendar>(value: DateTime Nullable) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<DateTime Nullable>(Calendar.SelectedDateProperty, value, ValueNone)
```
there were nullable dates in the Calendar/DatePicker properties.
I Initialy tried to go for the `DateTime option` but when the app ran it crashed due to the datetimes not being nullable so I switched to `DateTime Nullable`. Not sure if that is a good way to handle it or there are other ways to deal with nullable types

I'd be glad to update in case there's a better approach

datepicker


src/Avalonia.FuncUI.DSL/Calendar/Calendar.fs
```diff
@@ -0,0 +1,64 @@
...
+        static member headerBackground<'t when 't :> Calendar>(value: IBrush) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<IBrush>(Calendar.HeaderBackgroundProperty, value, ValueNone)
+
+        static member displayMode<'t when 't :> Calendar>(value: CalendarMode) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<CalendarMode>(Calendar.DisplayModeProperty, value, ValueNone)
+
+        static member selectionMode<'t when 't :> Calendar>(value: SelectionMode) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<SelectionMode>(Calendar.SelectionModeProperty, value, ValueNone)
+
+        static member selectedDate<'t when 't :> Calendar>(value: DateTime Nullable) : IAttr<'t> =
```
JaggerJo: include overloads (is usually done for `Nullable` in FuncUI)
see
https://github.com/AvaloniaCommunity/Avalonia.FuncUI/blob/1ac7ce77a9ff0865755b9d6c667ab145b801f16a/src/Avalonia.FuncUI.DSL/ToggleButton.fs#L23

JaggerJo: ohh, I've just seen you already did this

JaggerJo:
Just a few more things I just thought about:

- Should bindings for `CalendarButton`, `CalendarDayButton` & `CalendarItem` be included ?

looks like most of their functionality is marked internal and I'm not sure if including bindings would be a good thing. (not sure if they would be ever used)

- Is it possible to select multiple dates and actually get the range ?

for me it looks like the DatePicker/Calendar implementation in Avalonia is a bit messy. Maybe some improvements to it could help funcUI to integrate better if this currently does not work.

What do you think ?

AngelMunoz:
>Should bindings for CalendarButton, CalendarDayButton & CalendarItem be included ?

Yup you're right, I thought these were other kind of UI views (instead of the textbox + button) but after a closer look I agree, I'll take them out

>Is it possible to select multiple dates and actually get the range ?

It doesn't look like the date picker allows for a range, the closes thing I've seen is on the Calendar which has a `BlackoutDates`  .net property that takes a `CalendarBlackoutDatesCollection` which I didn't include a binding since it seems it's read only
it would certainly need to be updated in Avalonia, I took a look at the issues and I couldn't see anything related to date picker Date Range

AngelMunoz:
Looks like the calendar element provides support for date ranges but in the date picker code it's set to  single mode
https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Calendar/DatePicker.cs#L429

JaggerJo:
Ok, thats also what I found.

https://github.com/AvaloniaUI/Avalonia/blob/d69e815a84d6a0721dcb7e8afa4ec7f94d376c22/src/Avalonia.Controls/Calendar/Calendar.cs#L628

This is fine for now I think. Thanks for the contribution, this was on my TODO list for a long time.

PR #67, opened 2020-01-30
Cleaned the code

Thecentury:
Hi @JaggerJo!
Please take a look at this clean up:
* Removed unused namespace imports
* Removed unnecessary `new`s
* When parameter is not used, added a `_` prefix to it (Rider/Resharper shows a warning saying that argument is not used if it doesn't begin from underscore)
* Added comments about suspicious code in VirtualDom.Patcher

Code in https://github.com/AvaloniaCommunity/Avalonia.FuncUI/compare/master...Thecentury:code-cleanup?expand=1#diff-d5dc7d50202468ce9b47b818c5cce381R111 looks suspiciously to me, is it correct?

### DIFF
--- a/src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/CanvasDemo.fs
+++ b/src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/CanvasDemo.fs
@@ -1,16 +1,13 @@
 ﻿namespace Avalonia.FuncUI.ControlCatalog.Views
 
 open Avalonia
-open System
 open Avalonia.Controls
 open Avalonia.Controls.Shapes
 open Avalonia.FuncUI.DSL
 open Avalonia.FuncUI.Components
 open Avalonia.FuncUI.Elmish
 open Elmish
-open Avalonia.Layout
 open Avalonia.Media
-open SharpDX.Direct3D11
 
 module CanvasDemo =
     type State = { itemCount: int }
@@ -24,7 +21,7 @@ module CanvasDemo =
         match msg with
         | SetCount count -> { state with itemCount = count }
            
-    let view (state: State) (dispatch) =
+    let view (_state: State) (_dispatch) =
         Canvas.create [
             Canvas.width 300.0
             Canvas.height 400.0
--- a/src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs
+++ b/src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs
@@ -15,7 +15,7 @@ module internal rec Patcher =
         let subscriptions =
             match ViewMetaData.GetViewSubscriptions(view) with
             | null ->
-                let dict = new ConcurrentDictionary<_, _>()
+                let dict = ConcurrentDictionary<_, _>()
                 ViewMetaData.SetViewSubscriptions(view, dict)
                 dict
             | value -> value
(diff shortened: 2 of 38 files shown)


src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/CanvasDemo.fs
```diff
@@ -24,7 +21,7 @@ module CanvasDemo =
         match msg with
         | SetCount count -> { state with itemCount = count }
            
-    let view (state: State) (dispatch) =
+    let view (_state: State) (_dispatch) =
```
JaggerJo: is this common practice somehow ? 

It would also be possible to discard both arguments to clarify that they are unused. 

https://sharplab.io/#v2:DYLgZgzgNAJiDUAfA9gBwKYDsAEBlAnhAC7oC2AsAFBWnIwCuw62AgtgLxVXY/ZNHYAhjBjYAFAH0QAS0xEo2KbKIBKbCGzKO3XroAMQA===

src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs
```diff
@@ -108,20 +108,27 @@ module internal rec Patcher =
             if List.isEmpty delta then
                 () // list is empty by default
             else
+                // index is not modified here,
```
JaggerJo: yep, something went wrong here. good catch.

src/Examples/Examples.GameOfLife/Main.fs
```diff
@@ -21,16 +18,6 @@ module Main =
     | StartEvolution
     | StopEvolution
 
-    let timer (state: State) =
```
JaggerJo: I guess that's by mistake ?

JaggerJo:
Hey @Thecentury!

Thanks for the PR.

> Code in https://github.com/AvaloniaCommunity/Avalonia.FuncUI/compare/master...Thecentury:code-cleanup?expand=1#diff-d5dc7d50202468ce9b47b818c5cce381R111 looks suspiciously to me, is it correct?

No it's not. Seems like nearly every collection implements List so no-one noticed it yet. (Control catalog runs fine).

This is definitely something that should be fixed and that could use some tests.

JaggerJo:
sent you a PR that should fix the IEnumerable bug and generally makes this a bit simpler

src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/CanvasDemo.fs
Thecentury: Yes, it is a common practice. 
For example, take a look at https://github.com/dotnet/fsharp/blob/master/src/fsharp/FSharp.Core/Linq.fs#L168.

I decided not to replace the whole arg name with '_' to save the meaning of arguments.

src/Examples/Examples.GameOfLife/Main.fs
Thecentury: It is not used here, the same code is defined in `Program.fs`.

src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/CanvasDemo.fs
JaggerJo: ok

src/Examples/Examples.GameOfLife/Main.fs
JaggerJo: ok

PR #69, opened 2020-02-04
Default value for Styles, Classes and Resources properties

Thecentury:


src/Avalonia.FuncUI.UnitTests/VirtualDom/VirtualDom.DifferTests.fs
```diff
@@ -353,6 +370,7 @@ module DifferTests =
                     Delta.AttrDelta.Property {
                         accessor = Accessor.AvaloniaProperty StackPanel.OrientationProperty
                         value = Some (Orientation.Vertical :> obj)
+                        defaultValueFactory = ValueNone
```
JaggerJo: might be a dumb question but why is it a factory ? I see that it could help in some cases where you need to create an object on demand, but not sure if that's actually a case.

JaggerJo:
Looks good to me. Thanks for the PR.

src/Avalonia.FuncUI.UnitTests/VirtualDom/VirtualDom.DifferTests.fs
Thecentury: @JaggerJo Initially I thought that it should be a factory, as `Classes` or `Styles` are mutable, and so the default value, stored in the property, can be changed. 
But after your question I realised that when property is updated, default value will be taken from new property, where it was not changed.

If the latter is true, I will change factory to just plain default value.

JaggerJo: In practice the `AttrBuilder` will act as a factory most of the time, but in some cases your factory design is actually needed. I had to give it some more thoughts and if views are cached then not having a factory could lead to reusing the same instance.

Let's keep it a factory. Good work, I would have missed that.

PR #70, opened 2020-02-05
add Presso Example

JaggerJo:


JaggerJo:
image

Simple app to calculate the coffee you need depending on the amount of water, brewing method and desired strength.

Hopefully this helps to add useful extensions to FuncUI, as it's a visually more complex example. While building this I already noticed there are some missing bindings we should definitely add (`clip` for example).

PR #74, opened 2020-02-09
Add Examples.MediaPlayer

AngelMunoz:
I cleaned up the sample as well as leaving it in a single project
it does not have anything related to databases or such.

I'm pretty sure it builds on windows, would you mind to check/add what's needed to build on mac please?


JaggerJo:
nice example :)

PR #75, opened 2020-02-09
Mediaplayer macOS VLC  native support and view format

JaggerJo:
Hey @AngelMunoz,

hope you don't mind that I reformatted the views. No idea why but I really prefer this formatting.


AngelMunoz:
Yeah, I don't mind. I'm using vscode with ionide which uses [fantomas](https://github.com/fsprojects/fantomas) for the formatting, I try to switch to visual studio when working/sending stuff to keep the formatting consistent to yours
but this came from the other project, that's completely formatted 😅 forgot to change the formatting

PR #79, opened 2020-03-03
Fixed sync dispatch

Thecentury:
CheckAccess works all the way around. Access check should be after receiving the message as we don't know from which source we got it.


Thecentury:
@JaggerJo Hi Josua, please take a look at this PR.

JaggerJo:
Thanks for the contribution!

Do you need a published version ?

BTW. What are you using FuncUI for (if you can/want to tell) ?

Thecentury:
> Do you need a published version ?
No, thank you, I have this fix in my code.
 
> BTW. What are you using FuncUI for (if you can/want to tell) ?
It's a small application, where user logs in, scans several barcodes, application moves him through the steps of scanning different products, then shows a summary and send scanned data to a processing.
I wanted to write an MVU GUI app, stumbled upon FuncUI and liked it.

PR #81, opened 2020-03-09
Relaxed the type of view func

Thecentury:
Now it allows for the viewFunc to return just an IView, which can be useful when e.g. viewFunc returns two different control types depending on the condition.


Thecentury:
@JaggerJo Hi Joshua, I've fixed the build. Code became a bit more difficult now, but at allows to reach the goal.

JaggerJo:
LGTM, Thanks for the contribution!

PR #83, opened 2020-03-13
Two subscriptions considered different if either of them captures state

Thecentury:
Fixed the issue with two subscriptions that capture state to be equal.

### DIFF
--- a/src/Avalonia.FuncUI/Types.fs
+++ b/src/Avalonia.FuncUI/Types.fs
@@ -75,11 +75,13 @@ module Types =
                 match other with
                 | :? Subscription as other -> 
                     this.name = other.name &&
-                    this.funcType = other.funcType
+                    this.funcType = other.funcType &&
+                    not this.funcCapturesState &&
+                    not other.funcCapturesState
                 | _ -> false
                 
             override this.GetHashCode () =
-                (this.name, this.funcType).GetHashCode()
+                (this.name, this.funcType, this.funcCapturesState).GetHashCode()
                                
     type IAttr =
         abstract member UniqueName : string
--- a/src/Avalonia.FuncUI.UnitTests/LibTests.fs
+++ b/src/Avalonia.FuncUI.UnitTests/LibTests.fs
@@ -1,6 +1,9 @@
 ﻿namespace Avalonia.FuncUI.UnitTests
 
+open System
+open System.Threading
 open Avalonia.FuncUI.Library
+open Avalonia.FuncUI.Types
 open Xunit
 
 module Library =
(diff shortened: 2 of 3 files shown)


Thecentury:
Hi @JaggerJo, I don't know whether there are notifications from Github PRs, so letting you know about this PR just in case.

JaggerJo:
@Thecentury Thanks for the PR. I hope I will find time to look at this tomorrow. (BTW. did not get a notification - normally I check every day but I'm currently moving)

src/Avalonia.FuncUI/Types.fs
```diff
@@ -75,11 +75,13 @@ module Types =
...
                     this.name = other.name &&
-                    this.funcType = other.funcType
+                    this.funcType = other.funcType &&
+                    not this.funcCapturesState &&
+                    not other.funcCapturesState
                 | _ -> false
                 
             override this.GetHashCode () =
-                (this.name, this.funcType).GetHashCode()
+                (this.name, this.funcType, this.funcCapturesState).GetHashCode()
```
JaggerJo: looks good

src/Avalonia.FuncUI/Types.fs
```diff
@@ -75,11 +75,13 @@ module Types =
                 match other with
                 | :? Subscription as other -> 
                     this.name = other.name &&
-                    this.funcType = other.funcType
+                    this.funcType = other.funcType &&
+                    not this.funcCapturesState &&
```
JaggerJo: not sure if this is the right thing here.

In theory if a function that captures state should be of a different type than a function that doesn't (Because the compiler will generate a different class as far as I know). 

With this equals implementation two subscriptions that are basically equal will return false.

```fsharp
let sub = {
    Subscription.name = "sub"
    subscribe = fun _ -> new CancellationTokenSource()
    func = dlgt
    funcCapturesState = true
    funcType = typeof<int>
}

sub.Equals(sub) // this returns false
```

> x.Equals(x) returns true.
[microsoft docs](https://docs.microsoft.com/en-us/previous-versions/ms173147(v=vs.90)?redirectedfrom=MSDN)

JaggerJo: Sorry that I haven't merged your PR and also did not came up with a fix myself. I'm currently rethinking the whole subscription model and am trying to come up with a better implementation that doesn't require to patch basically all subscriptions on every view update.

There is another issue with subscriptions that leads to a loop of (patch -> render -> patch -> ..) because presubscription of a property invokes the subscription once.

Thecentury: I think here https://github.com/Thecentury/Avalonia.FuncUI/commit/07df113485870ea51a7006093040393ddff1d741 in two subsequent calls of the view function `funcType` was the same, otherwise there would be no problem.

Thecentury: >  I'm currently rethinking the whole subscription model

Nice to hear that :)

My fix is indeed a shallow one.

PR #88, opened 2020-03-18
New concept subscription handling

JaggerJo:


JaggerJo:
Hey @Thecentury,

could you take a look ?

This is not a ready to merge PR at all, but implements a new concept of determining when a subscriptions need to be patched.

# How ?:
Every Subscription now has a `scope` field (need to find a better name) and if that changes the subscription needs to be patched. Thats it.

```fsharp
// will still work, but not be patched
Button.onClick (fun _ -> 
    // do something
)
```

```fsharp
Button.onClick (fun _ -> 
    // do something and use the current state
, state) // <- will be patched if the state changed
```

```fsharp
Button.onClick (fun _ -> 
    // do something and use the current state
, Guid.NewGuid()) // <- will be patched every time (there should be a DU that maps the cases.)
```

Why I think this is worth considering:
- makes diffing simpler, no special `ForcePatch` and `FunctionAnalysis` anymore.
- makes dependencies in subscriptions explicit
- reduces required patching

JaggerJo:
@AngelMunoz what do you think ?

Thecentury:
> // will still work, but not be patched
Button.onClick (fun _ -> 
    // do something
)

If you remove analysis of whether function captures the state, can this kind of usage of subscriptions introduce a potentially incorrect behavior when function will actually use the state, but now subscription will not be patched?

> Button.onClick (fun _ -> 
    // do something and use the current state
, state) // <- will be patched if the state changed

Will `box` operator be needed here as `state` has a type of `obj`?

Could the callback function accept not only event args but a state too to make it not capture everything outside? 

> Button.onClick (fun _ -> 
    // do something and use the current state
, Guid.NewGuid()) // <- will be patched every time (there should be a DU that maps the cases.)

Could it be worth to introduce one special `obj` value like `Subscription.ChangesAlways` to eliminate the need to pass a random value here? (My naming is just to express my idea). But this approach could potentially lead to a situation when `x.Equals(x)` will be `false`.

JaggerJo:
> I like the idea of this change.

Good to hear :)

> > // will still work, but not be patched
> > Button.onClick (fun _ ->
> > // do something
> > )
> 
> If you remove analysis of whether function captures the state, can this kind of usage of subscriptions introduce a potentially incorrect behavior when function will actually use the state, but now subscription will not be patched?
> 
> > Button.onClick (fun _ ->
> > // do something and use the current state
> > , state) // <- will be patched if the state changed
> 
> Will `box` operator be needed here as `state` has a type of `obj`?

not sure what you mean here. 
ValueTypes will be boxed.

> Could the callback function accept not only event args but a state too to make it not capture everything outside?

Would be really complicated to archive. The VirtualDom has no idea there is something like a state (and I think this is good).

> > Button.onClick (fun _ ->
> > // do something and use the current state
> > , Guid.NewGuid()) // <- will be patched every time (there should be a DU that maps the cases.)

> Could it be worth to introduce one special `obj` value like `Subscription.ChangesAlways` to eliminate the need to pass a random value here? (My naming is just to express my idea). But this approach could potentially lead to a situation when `x.Equals(x)` will be `false`.

Good question, thought about this a lot. 

I thought we could expose a DU in the DSL and then later on just create a new Guid in the builder for the `Subscription.ChangeAlways` case.

AngelMunoz:
I have no idea 😄 or to be more accurate I don't understand the underlying needs for the change 
(and In general I don't understand much of the core code fo FuncUI I haven't had the time to dive into it)
I do feel like introducing DU like `Subscription.ChangesAlways` as @Thecentury mention is a great idea though :)

Thecentury:
@JaggerJo I will try to better explain my idea.

Before these changes FuncUI did some analysis of the subscription callback to determine whether it captures the state or not, and if it did then engine forcibly patched the subscription (except the cases like in my repro when two subscriptions were considered equal not taking their `capturesState` attribute into account).

As I understood, now you want to remove this analysis and instead give user an ability to pass the state manually, and FuncUI will decide whether to patch the subscription based on the change of this state.

But users would be able to still use the existing overload of a subscription without passing the state, and then their subscriptions will not be patched anymore as you removed the "captures state" analysis, so they will get a hard-noticeable bug, that they will be able not to catch at the first glance.
Moreover, when users who used the previous versions of the package will update it to a newer one, even without a desire to get a new subscriptions patching functionality, but to get other new features, the behavior of their programs will change, and as I see it, it will even break as subscriptions will stop to update.

Please correct me if I'm wrong.

JaggerJo:
> @JaggerJo I will try to better explain my idea.
> 
> Before these changes FuncUI did some analysis of the subscription callback to determine whether it captures the state or not, and if it did then engine forcibly patched the subscription (except the cases like in my repro when two subscriptions were considered equal not taking their `capturesState` attribute into account).
> 
> As I understood, now you want to remove this analysis and instead give user an ability to pass the state manually, and FuncUI will decide whether to patch the subscription based on the change of this state.
> 
> But users would be able to still use the existing overload of a subscription without passing the state, and then their subscriptions will not be patched anymore as you removed the "captures state" analysis, so they will get a hard-noticeable bug, that they will be able not to catch at the first glance.

> Moreover, when users who used the previous versions of the package will update it to a newer one, even without a desire to get a new subscriptions patching functionality, but to get other new features, the behavior of their programs will change, and as I see it, it will even break as subscriptions will stop to update.
> 
> Please correct me if I'm wrong.

Thats right. Another option would be to change the default behaviour to `Subscription.ChangeAlways`, but this will push people in a wrong direction (it's harder to do the right thing - bad defaults).

If it'a a breaking change this should be made explicit, you're right.

We could make a new the parameter required and leave the old functions but mark them as obsolete. Later on they could be removed and the parameter could be marked optional, but this only works if people migrate to that "intermediate" version.

What would you suggest ?

Thecentury:
@JaggerJo At the end what will be the behavior of the subscription when no `scope` is provided?
Will it be never patched?
If so, could the `scope` parameter made required. To create a never-patched subscription users will provide some other constant special-case value like `NeverPatch`.
So eventually there current overload of subscriptions will be removed, thus removing the possibility to use subscriptions incorrectly.

What do you think?

JaggerJo:
Will merge this if I don't receive other feedback. @Thecentury thumbs up pls ;)

Thecentury:
@JaggerJo Looks good!

bradphelan:
Can you please provide some use cases explaining what might happen if you don't use this correctly. Previously the concept was simple. **view** is a pure function and returns a description of the UI. The real UI reflects what is returned by the view. However under the hood magic happens to improve performance. 

Funny...I started googling "comparing functions for equality" and I found https://www.reddit.com/r/fsharp/comments/btpc4h/comparing_f_functions_for_equality/ which is your question @JaggerJo  to the FSharp community.

Is it the case that it is just plain impossible to correctly diff a function in F#?

PR #91, opened 2020-03-26
Revert "Fix Media bar in Player Sample"

AngelMunoz:
Reverts AvaloniaCommunity/Avalonia.FuncUI#90


AngelMunoz:
it seems that I didn't merge the PR before starting to work on the docs stuff, so the partial stuff leaked in 😭 my bad

PR #96, opened 2020-03-28
[WIP] created simple `useEffect` prototype

JaggerJo:


Zaid-Ajaj:
Hello @JaggerJo, this feature seems interesting. However, if you are aiming for React-like implementation of `useEffect`, then this very far from it. In fact, this usage of `useEffect` has nothing in common with the one from React. I encourage to change the name into something else to avoid confusion. 

Gaining access to the internal reference of the control is closer what the [ref](https://reactjs.org/docs/refs-and-the-dom.html) attribute represents in the React world. 

Effects in React are shorthand for controlling the life-cycle events (componentDidMount, componentDidUnmount) of stateful components where each components has its own state internally. AFAIK this library has the concept of components because in pure MVU world, these isn't one.

JaggerJo:
> Hello @JaggerJo, this feature seems interesting. However, if you are aiming for React-like implementation of `useEffect`, then this very far from it. In fact, this usage of `useEffect` has nothing in common with the one from React. I encourage to change the name into something else to avoid confusion.

Hey @Zaid-Ajaj, thanks for the Feedback.  👍

The idea of **$featureName** is to make it easier to get a control reference and do something with it (that is not tracked). Currently the idea is to provide options to do this **on create** (not sure if on control creation or on visual/logical tree attach), **on destruct** (same here basically) and on every state update. 

> Gaining access to the internal reference of the control is closer what the [ref](https://reactjs.org/docs/refs-and-the-dom.html) attribute represents in the React world.

Hmm, yes basically it gives you a way of accessing the internal reference and doing things with it.

> Effects in React are shorthand for controlling the life-cycle events (componentDidMount, componentDidUnmount) of stateful components where each components has its own state internally. AFAIK this library has the concept of components because in pure MVU world, these isn't one.

Seems also applicable, right ? In Avalonia you sometimes also want to just call a method on a control. It's not the norm, but there are legitimate uses of it.

Do you have a better name suggestion for the feature ? 😀 naming things is hard sometimes..

src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/MaskedTextBoxDemo.fs
```diff
@@ -0,0 +1,72 @@
...
+            StackPanel.spacing 10.0
+            StackPanel.children [
+                TextBox.create [
+                    TextBox.text (string state.count)
+                    TextBox.onTextChanged (fun text ->
+                        let isNumber, number = Int32.TryParse text
+                        if isNumber then
+                            dispatch (SetCount number)
+                            
+                    , SubPatchOptions.Never) // <- this is annoying.
```
AngelMunoz: Yeah, I think we should open a separate issue to track things like these that need to be documented and go for some defaults

Zaid-Ajaj:
> Do you have a better name suggestion for the feature ? 😀 naming things is hard sometimes..

"Control life-cycle events" sounds appropriate to me :smile: I like `onCreate` and `onDestroy` as well

src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/MaskedTextBoxDemo.fs
```diff
@@ -0,0 +1,72 @@
...
+                        
+                        let action = EventHandler<TextInputEventArgs>(fun sender args ->
+                            let isNumber, _ = Int32.TryParse args.Text
+                            if not isNumber then
+                                args.Text <- String.Empty
+                        )
+                            
+                        (TextBox.TextInputEvent, action, RoutingStrategies.Tunnel, false)
+                        |> control.AddHandler
+                        |> ignore
```
AngelMunoz: this kind of usage could lead to memory leaks if it goes unchecked right? It would be nice to add something in the docs about it

JaggerJo: Yes, this leaks like crazy :D 

It's just as a prove of concept - later this would be only done once, in the **$onCreate**. This was just a test to see if this would actually work in order to bypass default event routing   and.. it does 😁

JaggerJo: (where in this case we basically intercept events in the `Tunneling` phase)

PR #100, opened 2020-03-31
Fixed DataTemplateView.Match method

Thecentury:
@JaggerJo Fixes https://github.com/AvaloniaCommunity/Avalonia.FuncUI/issues/99


JaggerJo:
Good catch! Thanks for the Contribution.

PR #102, opened 2020-03-31
[WIP] Mockup of attr with list of attrs

Thecentury:
Hi @JaggerJo, what do you think about this idea?

It gives an ability to create custom methods returning a list of connected attrs — e.g., list with two subscriptions for `onChecked` and `onChecked` events,
or static-typed setting of both `ItemsControl.Items` and `ItemsTemplate`, where the type of an element of the `Items` gives the data type in the template.

Or any other example of creating a list of attributes from one piece of data.

### DIFF
--- a/src/Avalonia.FuncUI/Builder.fs
+++ b/src/Avalonia.FuncUI/Builder.fs
@@ -216,9 +216,11 @@ type AttrBuilder<'view>() =
 [<AbstractClass; Sealed>] 
 type ViewBuilder() =
     
-    static member Create<'view>(attrs: IAttr<'view> list) : IView<'view> =
+    static member Create<'view>(attrs: Attr<'view> list) : IView<'view> =
+        let innerAttrs = attrs |> List.collect Attr.toInnerAttrs
+
         {
             View.viewType = typeof<'view>
-            View.attrs = attrs
+            View.attrs = innerAttrs
         }
         :> IView<'view>
--- a/src/Avalonia.FuncUI/Types.fs
+++ b/src/Avalonia.FuncUI/Types.fs
@@ -82,13 +82,13 @@ module Types =
                 
             override this.GetHashCode () =
                 (this.name, this.funcType, this.scope).GetHashCode()
-                               
+
     type IAttr =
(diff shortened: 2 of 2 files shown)


Thecentury:
Also it will allow to set no attributes conditionally, represented by a list with no elements:

```fsharp
let setSmth param =
  if param then
    Button.setSmth 1
  else
    Button.attrList []

Button.create [
  Button.content "Click"
]
```

If is possible to achieve the same now using `yield! some_possibly_empty_list_of_attributes`, but it has more verbose syntax.

JaggerJo:
Hey @Thecentury,

sorry for the late response. 

If I understand your proposal correctly it basically enables you to add a list of attributes in the declaration, right ?

```fsharp

let reusableGroupOfAttributes : IAttr<Button> list =
    [
        Button.verticalAlignment VerticalAlignment.Center
        Button.horizontalAlignment HorizontalAlignment.Center
        Button.margin 20.0
    ]

let wrap (attrs: IAttr<'t> list) : 


Button.create [
    Button.background "green"
    Button.content "click me"
]
```

Thecentury:
Hi @JaggerJo,

I already added something similar to `View.withAttrs` to my code, it works, but it looks not as setting properties in the native FuncUI way, a bit artificially.

I think, the change I propose can be shallow — we can create new type for inner attributes with the same set of possible values (property, content and subscription), and all implementation of Virtual DOM will remain unchanged. Conversion will happen at the edge, just in function creating view with attributes, as in my example.

I will try to implement this feature as a wrapper around existing API, just to check whether it is possible.

Thecentury:
Hi @JaggerJo, should I close this PR?

JaggerJo:
Hey @Thecentury , Yes sure. feel free to close it. Are you ok with the current solution ?

Thecentury:
@JaggerJo Yes, I left the current solution. This idea was an improvement, and as I switched to another project I didn't have enough time and motivation to work on it.

JaggerJo:
@Thecentury Let's keep this idea in mind. We might need to implement something like this in the future. 🙂

PR #110, opened 2020-04-05
Update Docs

AngelMunoz:
This PR adds DatePicker and Checkbox doc files


src/docs/controls/CheckBox.md
```diff
@@ -0,0 +1,66 @@
...
+CheckBox.create [
+    // can be either true or false
+    Checkbox.isChecked state.booleanValue
+]
+```
+
+**Set Indeterminate**
+```fsharp
+CheckBox.create [
+    // can be either true or false
```
JaggerJo, line 37: `true` / `false` or `null` right ?

AngelMunoz: I played a little bit with this yesterday it seems you need to have both `isThreeState` in true and `isChecked` in null (None or nullable bool set to null) for it to work  or else it will show as unselected (`isChecked` false) it's a bit weird actually

PR #118, opened 2020-04-30
Update Avalonia to 0.9.9

FoggyFinder:
I've noticed that `MusicPlayer` has stale reference of Avalonia (`0.9.2` ) that leads to a crash on start so I decided to update version there and elsewhere as well.


JaggerJo:
Nice work 👍 Thanks for the contribution

Issue #146, opened 2020-12-31
Any plans to support DataGrid?


schwasam:
Hi,

Are there any plans to support DataGrid?
I understand that it was experimental/alpha in Avalonia at some point, but is now available as a separate NuGet package (Avalonia.Controls.DataGrid).

Or what would be the alternative to use right now?

JaggerJo:
Hey, 

Hmm, good question. When this was considered the last time there was no maintainer for the DataGrid implementation on the avalonia side and the API wasn't final IIRC.

I'm happy to take PR's that adds bindings for the DataGrid Control.

AngelMunoz:
by the way, we also added a "Guide" on how to create bindings a few days ago
https://avaloniacommunity.github.io/Avalonia.FuncUI.Docs/guides/Bindings.html

In case someone wants to take a jab at this

artipo:
I was looking into this issue, but saw @albert-du PR. What should we do?
In case `DataGrid` should not be included in Avalonia.FuncUI maybe this issue should be closed.

Thanks.

schwasam:
Yes from reading the PR comments it looks like DataGrid doesn't work well with F#/immutability.
Closing it.

PR #153, opened 2021-02-11
Add DataGrid Bindings

albert-du:
* Added Bindings for DataGrid #146 
* Added Package reference `<PackageReference Include="Avalonia.Controls.DataGrid" Version="0.10.0"`
* Added Control Catalog sample
  *  Requires a style reference `this.Styles.Load "avares://Avalonia.Controls.DataGrid/Themes/Fluent.xaml"`

The DataGrid items require `[<CLIMutable>]` for editing.
I haven't found a way to remove items or update the state with new items

### DIFF
--- a/src/Avalonia.FuncUI.ControlCatalog/Avalonia.FuncUI.ControlCatalog.fsproj
+++ b/src/Avalonia.FuncUI.ControlCatalog/Avalonia.FuncUI.ControlCatalog.fsproj
@@ -26,6 +26,7 @@
         <Compile Include="Views\Tabs\SplitViewDemo.fs" />
         <Compile Include="Views\Tabs\TickBarDemo.fs" />
         <Compile Include="Views\Tabs\WindowMenuDemo.fs" />
+        <Compile Include="Views\Tabs\DataGridDemo.fs" />
         <Compile Include="Views\MainView.fs" />
         <AvaloniaResource Include="**\*.xaml" />        
         <Compile Include="Main.fs" />
--- a/src/Avalonia.FuncUI.ControlCatalog/Main.fs
+++ b/src/Avalonia.FuncUI.ControlCatalog/Main.fs
@@ -26,6 +26,7 @@ type App() =
     override this.Initialize() =
         this.Styles.Add (FluentTheme(baseUri = null, Mode = FluentThemeMode.Dark))
         this.Styles.Load "avares://Avalonia.FuncUI.ControlCatalog/Styles/TabControl.xaml"
+        this.Styles.Load "avares://Avalonia.Controls.DataGrid/Themes/Fluent.xaml"
 
     override this.OnFrameworkInitializationCompleted() =
         match this.ApplicationLifetime with
--- a/src/Avalonia.FuncUI.ControlCatalog/Views/MainView.fs
+++ b/src/Avalonia.FuncUI.ControlCatalog/Views/MainView.fs
@@ -102,5 +102,9 @@ module MainView =
                     TabItem.header "TickBar Demo"
                     TabItem.content (ViewBuilder.Create<TickBarDemo.Host>([]))
                 ]
+                TabItem.create [
+                    TabItem.header "DataGrid Demo"
+                    TabItem.content (ViewBuilder.Create<DataGridDemo.Host>([]))
+                ]
             ]
         ]
--- a/src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/DataGridDemo.fs
+++ b/src/Avalonia.FuncUI.ControlCatalog/Views/Tabs/DataGridDemo.fs
@@ -1,0 +1,59 @@
+namespace Avalonia.FuncUI.ControlCatalog.Views
+
+open System
+open Avalonia.Controls
+open Elmish
+open Avalonia.FuncUI.DSL
+open Avalonia.FuncUI.Components
+open Avalonia.FuncUI.Elmish
+
+module DataGridDemo =
+    // [<CLIMutable>] required for editability
+    [<CLIMutable>] 
+    type Product =
+        { Name : string
+          Price : string
+          FavoriteColor : string
+          Description : string }
+        with
+            static member Random () : Product =
+                let faker = Bogus.Faker(locale = "de")
+                { Name = faker.Commerce.Product()
+                  Price = faker.Commerce.Price(0.99m, 1000.0m, 2)
+                  FavoriteColor = faker.Random.Hexadecimal(6, "#")
+                  Description = faker.Lorem.Sentences(Nullable(3)) }
+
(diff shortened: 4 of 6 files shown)


AngelMunoz:
thank you for sending this PR give us a few days or even a couple of weeks to review this change we appreciate it

albert-du:
I've been unable to make the DataGrid Editable through without allowing mutation.
The problem is in Avalonia.Controls.DataGrid
```csharp
public bool GetPropertyIsReadOnly(string propertyName)
        {
            if (DataType != null)
            {
                if (!String.IsNullOrEmpty(propertyName))
                {
                    Type propertyType = DataType;
                    PropertyInfo propertyInfo = null;
                    List<string> propertyNames = TypeHelper.SplitPropertyPath(propertyName);
                    for (int i = 0; i < propertyNames.Count; i++)
                    {
                        propertyInfo = propertyType.GetPropertyOrIndexer(propertyNames[i], out object[] index);
                        if (propertyInfo == null || propertyType.GetIsReadOnly() || propertyInfo.GetIsReadOnly())
                        {
                            // Either the data type is read-only, the property doesn't exist, or it does exist but is read-only
                            return true;
                        }

                        // Check if EditableAttribute is defined on the property and if it indicates uneditable
                        object[] attributes = propertyInfo.GetCustomAttributes(typeof(EditableAttribute), true);
                        if (attributes != null && attributes.Length > 0)
                        {
                            EditableAttribute editableAttribute = attributes[0] as EditableAttribute;
                            Debug.Assert(editableAttribute != null);
                            if (!editableAttribute.AllowEdit)
                            {
                                return true;
                            }
                        }
                        propertyType = propertyInfo.PropertyType.GetNonNullableType();
                    }
                    return propertyInfo == null || !propertyInfo.CanWrite || !AllowEdit || !CanEdit(propertyType);
                }
                else if (DataType.GetIsReadOnly())
                {
                    return true;
                }
            }
            return !AllowEdit;
        }
```
The property info for an f# record's property returns as false, setting `isReadOnly` to true
```csharp
!propertyInfo.CanWrite
```
https://github.com/AvaloniaUI/Avalonia/blob/94300c7d8b75f470ae1565d82cdb019645a37fab/src/Avalonia.Controls.DataGrid/DataGridDataConnection.cs#L368-L407

As far as I can tell, there is no way to force `isReadOnly` to false while using immutable f# records.
Setting the `isReadOnly` property through bindings won't solve this problem either, 
```csharp
        internal bool GetColumnReadOnlyState(DataGridColumn dataGridColumn, bool isReadOnly)
        {
            Debug.Assert(dataGridColumn != null);

            if (dataGridColumn is DataGridBoundColumn dataGridBoundColumn && 
                dataGridBoundColumn.Binding is BindingBase binding)
            {
                var path = (binding as Binding)?.Path ?? (binding as CompiledBindingExtension)?.Path.ToString();

                if (string.IsNullOrWhiteSpace(path))
                {
                    return true;
                }
                else
                {
                    return DataConnection.GetPropertyIsReadOnly(path) || isReadOnly;
                }
            }

            return isReadOnly;
        }
```
https://github.com/AvaloniaUI/Avalonia/blob/94300c7d8b75f470ae1565d82cdb019645a37fab/src/Avalonia.Controls.DataGrid/DataGridColumns.cs#L140-L160

Overall, Avalonia's DataGrid is extremely reliant on mutation and data binding for editing. While the possibility exists for including it as readonly, very little of its original functionality remains,  At this time I no longer think bindings for Avalonia's DataGrid should be added to FuncUI and I'll close this pr.

fischgeek:
Sorry to surface this, but is it at all possible to include it as a read-only somehow or am I not understanding/doing it correctly?

JordanMarr:
It seems to me like it would be worth having even as read only.

JaggerJo:
@fischgeek I think there a multiple ways of using a `DataGrid` without writing complicated declarative bindings for it. 

1. Embed a `DataGrid` in a component and wire things up manually by either:
- obtaining a control reference via `View.createWithOutlet`
- executing code in the new `Control.init (fun dataGrid -> ...)` attribute

2. Use a `StaticComponent` (experimental) and setup bindings Manually. This requires:
- creating minimal bindings for `DataGrid`
-setting bindings up manually

3. Create a normal Avalonia Control / View and just use that in FuncUI

Pretty sure there are more ways.

JaggerJo:
Also, it might very well be possible to create bindings for `DataGrid` now as both FuncUI changed to support diffing non view (but AvaloniaObject) types and Avalonia also went through significant changes.

fischgeek:
> @fischgeek I think there a multiple ways of using a `DataGrid` without writing complicated declarative bindings for it.
> 
>     1. Embed a `DataGrid` in a component and wire things up manually by either:
> 
> 
>     * obtaining a control reference via `View.createWithOutlet`
> 
>     * executing code in the new `Control.init (fun dataGrid -> ...)` attribute
> 
> 
>     2. Use a `StaticComponent` (experimental) and setup bindings Manually. This requires:
> 
> 
>     * creating minimal bindings for `DataGrid`
> 
>     * setting bindings up manually
> 
> 
>     3. Create a normal Avalonia Control / View and just use that in FuncUI
> 
> 
> Pretty sure there are more ways.

Forgive the level of understanding here (or lack thereof), but I'm not sure what you mean by those options. I believe what you say will work, but I'm having a difficult time implementing on my end. 

For example, how do I turn a DataGrid into an "outlet" so I can pass it into `View.createWithOutlet`? There doesn't seem to be a `Control.init`. 

I skipped 2 because it seems way over my head. 

As for 3, I did create a DataGrid like so `let dg = new Avalonia.Controls.DataGrid()` but not sure how to get the View out of it now and use it.

JordanMarr:
I think the new `Control.init` would work for you, but it looks like a new preview package needs to be released to NuGet first.

JordanMarr:
@JaggerJo, is the `Control.init` implementation ready to be merged so we can release it as a preview?

JaggerJo:
> @JaggerJo, is the `Control.init` implementation ready to be merged so we can release it as a preview?
> 
> 

Yes! IIRC it's already merged

JordanMarr:
> Great! How do I go about getting that updated?

You are good to go on preview 8!

fischgeek:
> You are good to go on preview 8!

Fantastic! Thank you. Though, I'm having a lot of difficulty getting it to compile so I can test it out. On top of that, not sure what I would do once I did get it running. I am happy waiting until an official release too. You don't need to spend your time holding my hand through this. However, having some documentation written up would be most helpful! And, again I'm happy waiting now that I know something is coming.

JordanMarr:
Preview 8 is on nuget:
https://www.nuget.org/packages/Avalonia.FuncUI/0.6.0-preview8

Issue #120, opened 2020-05-03
Improve list diffing/patching performance


JaggerJo:
FuncUI currently implements super simple list diffing that works just fine in most cases.

There are cases where a list contains complex or a lot of different items this can be problematic. Especially when items are inserted at index 1 - because this currently will result in rebuilding all items.

Implementing a 2nd list diffing strategy that utilises keys to reduce the required patch work could speed up list diffing by a lot. ([here](https://reactjs.org/docs/reconciliation.html#keys) is how react does it)

Adding a Key to Avalonia Controls is easily possible using AttachedProperties. (We actually implicitly attache properties to controls for internal subscription handling - IIRC)

```fsharp
StackPanel.children [
    TextBlock.create [
        TextBlock.key "item 1"
        TextBlock.text "item 1"
        ...
    ]
    ...
]
```

WillEhrendreich:
should this be open, still? the discussion on #157 seems to suggest it's been integrated into things, now.

PR #157, opened 2021-03-04
Feature: View key property

B-Reif:
This PR adds a new base property (and attr) to FuncUI called "key". FuncUI notes whether or not the key attribute changes in a ViewDelta. If the key has changed, then FuncUI will teardown the IControl and create a new one, instead of patching the existing IControl.

### Motivation
FuncUI currently decides whether or not to patch an IControl based only on its type. By exposing the "key" attribute, the app author can explicitly choose to re-create the IControl from scratch instead of diffing it. This can improve performance and behavior in cases when diffing is inappropriate. 

### Example
Suppose an app switches between different collections:

```fs
type ListKind = Letters | Numbers
type State = {
  Letters: char list
  Numbers: int list
  ActiveList: ListKind
}
let view state =
  match state.ActiveList with
  | Letters -> 
    ListBox.create [ 
      List.dataItems state.Letters
      // template goes here
    ]
  | Numbers -> 
    ListBox.create [ 
      List.dataItems state.Numbers
      // template goes here
    ]
```

This view function always returns a ListBox view. When ActiveList changes from Letters to Numbers, FuncUI will attempt to diff the contents of all of these lists. This diff would have a time complexity relative to both lists, and the change in the item type (from char to int) could also cause unpredictable behavior. By specifying a key attribute, we can tell FuncUI to skip diffing:

```fs
let view state =
  match state.ActiveList with
  | Letters -> 
    ListBox.create [ 
      Control.key "lettersList"
      // ...
    ]
  | Numbers -> 
    ListBox.create [
      Control.key "numbersList"
      // ... 
    ]
```

Now when the view function returns a ListBox view, FuncUI will notice the `key` attr has changed and skip diffing. This will improve rendering performance and improve behavior when changing item type.

In the future, we could also use the key attr to match views across renders and patch instead of re-create (for example, when the same elements are re-ordered). For now this PR only supports the "skip diffing" behavior.

### Prior Art
[Keys in React](https://reactjs.org/docs/lists-and-keys.html#keys)
[Reconciliation in React](https://reactjs.org/docs/reconciliation.html#recursing-on-children)

### See also
#120


src/Avalonia.FuncUI.DSL/Base/Control.fs
```diff
@@ -31,4 +33,5 @@ module Control =
         static member contextMenu<'t when 't :> Control>(menu: ContextMenu) : IAttr<'t> =
             AttrBuilder<'t>.CreateProperty<ContextMenu>(Control.ContextMenuProperty, menu, ValueNone)
            
-           
\ No newline at end of file
+        static member key<'t when 't :> Control>(key: string) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<string>(ViewMetaData.KeyProperty, key, ValueNone)
```
JaggerJo, line 37: Having UUIDs is common for items in a list - providing an overload would be handy.
```suggestion
        static member key<'t when 't :> Control>(key: string) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<string>(ViewMetaData.KeyProperty, key, ValueNone)
            
        static member key<'t when 't :> Control>(key: System.Guid) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<string>(ViewMetaData.KeyProperty, string key, ValueNone)
```

JaggerJo (approved):
Thanks for the contribution! 

Looks good so far. Would be nice to do some more clever things with list diffing and patching.

dsyme:
@B-Reif Are you interested in becoming co-maintainer on this repository?  Please see #172

JaggerJo:
Keys are implemented in my PR https://github.com/fsprojects/Avalonia.FuncUI/pull/179 

It turned out there is a solution that does not require iterating the attributes twice. 
Thanks for the work - only realised it after using your implementation.

PR #168, opened 2021-07-19
Fix some property bindings which fail at runtime, including Image.source

### LINKED ISSUE #167: Some property bindings are wrong in DSL
I noticed that `Image.source` failed at runtime because the Avalonia property is of type `IImage` not `IBitmap`.

I changed the property builder in `Builder.fs` to check that the types on the DSL bindings matched the types of the `AvaloniaProperty` they wrap.

That required a couple of extra binding methods, one to ensure that `ITemplate<IControl>` bindings would still build, and one to handle `string`-typed property bindings for Avalonia properties of type `obj`.

The summary of fixes is:
* `Image.source` was taking `IBitmap` instead of `IImage`
* `AutoCompleteBox.itemFilter` had the wrong signature for the search predicate
* `AutoCompleteBox.textFilter` was taking a `string` instead of a predicate
* `Visual.opacity` was attempting to bind to `IsVisible` instead of `Opacity` 
* `Visual.renderTransform` was of type `Transform` instead of `ITransform`
* same for `LayoutTransformControl.layoutTransform`
* `Decorator.child` was of type `obj` instead of `IControl`
* `Popup.placementAnchor` was attempting to bind to `PlacementMode` instead of `PlacementAnchor`
* `TabControl.selectedContent` was attempting to bind to `ContentTemplate` instread of `SelectedContent`

I'll submit a PR with these fixes (and the preventative changes). Hopefully I haven't broken anything in the process, the sample apps seemed to work OK, but I also don't think they use any of those properties.

marklam:
Fixes #167

### DIFF
--- a/src/Avalonia.FuncUI.DSL/AutoCompleteBox.fs
+++ b/src/Avalonia.FuncUI.DSL/AutoCompleteBox.fs
@@ -62,12 +62,12 @@ module AutoCompleteBox =
         static member filterMode<'t when 't :> AutoCompleteBox>(mode: AutoCompleteFilterMode) : IAttr<'t> =
             AttrBuilder<'t>.CreateProperty<AutoCompleteFilterMode>(AutoCompleteBox.FilterModeProperty, mode, ValueNone)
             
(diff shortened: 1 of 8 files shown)


JaggerJo (approved):
LGTM

PR #177, opened 2021-12-09
Avalonia upgrade to 0.10.10

sleepyfran:
This PR includes the upgrade of the `Avalonia.Desktop` package used in the main `Avalonia.FuncUI` project to the latest version (0.10.10)

Validated all example apps on macOS Monterey with .NET 5.0:
image
image
image
image
image


JordanMarr (commented):
Is there any reason to bump to the latest version of FSharp.Core? 
If not, it would probably be best to keep it at an older version to allow more backwards compatibility.

sleepyfran:
> Is there any reason to bump to the latest version of FSharp.Core? If not, it would probably be best to keep it at an older version to allow more backwards compatibility.

There's no real reason, just thought of upgrading to the latest versions since I was at it. But you're right, reverted the version back to `4.7.2`, there's only some formatting changes in the fsproj file, but other than that just Avalonia upgraded now 😄

JordanMarr (approved):

Issue #172, opened 2021-12-06
Seeking new co-maintainers for Avalonia.FuncUI


dsyme:
@JaggerJo is seeking new co-maintainers for Avalonia.FuncUI

This is an important project for the F# ecosystem and is listed as the No 1 option for "Use F# for Desktop Apps" in https://fsharp.org/use/desktop-apps/.  It is usable, stable and with low bug counts, and with good engineering practices.  

I'm interested in doing some backup co-maintenance myself, and will be keeping an eye on things. 

Please let us know if you're interested in becoming co-maintainer here.

Many thanks!

for @JaggerJo, @sergey-tihon and myself

sleepyfran:
Hey Don! Thanks for bringing this up and for the ping on the upgrade PR 😃 

I'd love to help with the maintenance, I'd probably not have time to be a full-on maintainer but I can definitely help with reviews, upgrades and some new features.

dsyme:
Fantastic!  I will make you co-maintainer - reviews, upgrades, looking after PRs etc is what we're after for sure

* Policies for fsprojects maintainers are [here](https://github.com/fsprojects/fsprojects-admin#an-f-community-project-incubation-space), take a quick glance, I think you know the gig

* You'll be added to the "maintainers" group and I'll send an announcement there

dsyme:
Cool thanks. We should likely move to auto-publish during main branch build (that's what FSharp.Data and several other "fsprojects" projects are doing)

dsyme:
And welcome @sleepyfran as a new co-maintainer! Great to have you on board :)

JaggerJo:
> > And welcome @sleepyfran as a new co-maintainer! Great to have you on board :)
> 
> Thank you Don! Will definitely take a look soon and update whatever's needed 😃

Nice! Happy to help if something comes up @sleepyfran. Automating releases seems like a good first step to simplify maintenance.

houstonhaynes:
I'd love to pitch in but I have quite a bit of woodshedding/headscratching to do in this department. I planned on using Avalonia for a WSL monitor project so it's right up my alley, but I won't be very useful until I've had some time with it. That said if you're willing to take on someone who needs a bit of "runway" to get acclimated I'd be happy to pitch in. 👍

unbyte:
Wow, I'm so glad to see this project has new maintainers! I gave up using F# to implement GUI before because I've encountered some problems and found that this library has fallen far behind the development progress of Avalonia.
[...]

PR #180, opened 2021-12-13
Automate package publication via GitHub Action

### LINKED ISSUE #178: Automate package publication
As mentioned in #172 the process of publishing new package versions is done manually. We should upgrade the process to be automatic as it's done in other `fsprojects`.

We can take the workflow in `FSharp.Data` as an example:
https://github.com/fsprojects/FSharp.Data/blob/main/.github/workflows/push-master.yml#L35

sleepyfran:
This PR closes #178 by creating a new workflow called `Build, Test and Publish release` which builds the library, runs all the UTs and publishes a new version to NuGet if the version has changed. Also took the opportunity to rename the old flow to `pull-request.yml` and `Build and test` and make it run only on new PRs to master. 

Heavily inspired by [FSharp.Data's flow](https://github.com/fsprojects/FSharp.Data/blob/main/.github/workflows/push-master.yml) and using the secret name that @dsyme provided in [here](https://github.com/fsprojects/Avalonia.FuncUI/issues/172#issuecomment-988945425). Validated with [Act](https://github.com/nektos/act).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,24 @@
+name: Build, Test and Publish release
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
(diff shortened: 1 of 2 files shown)


JordanMarr (approved):
Nice!

PR #182, opened 2022-01-10
ChordParser example with useElmish hook

JordanMarr:
ChordParser component example + an experimental `useElmish` hook (inspired by Feliz)!


src/Avalonia.FuncUI.Components/Context/Context.Hooks.fs
```diff
@@ -145,4 +145,15 @@ module IComponentContextExtensions =
...
+        /// Provides Elmish style updates for a writable model.
+        member this.useElmish<'Model, 'Msg> (writableModel: IWritable<'Model>, update) = 
+            let state = ElmishHook.ElmishState<'Model, 'Msg>(writableModel, update)
+            state.Model, state.Dispatch
+
+        /// Provides Elmish style updates for a model via the useState hook.
+        member this.useElmish<'Model, 'Msg> (init: 'Model, update) = 
+            let writableModel = this.useState (init, true)
+            let state = ElmishHook.ElmishState<'Model, 'Msg>(writableModel, update)
+            state.Model, state.Dispatch
```
JaggerJo, line 159: @JordanMarr it should be possible to move the Elmish related stuff to the FuncUI Elmish Project, right?

JaggerJo (commented):
LGTM

src/Avalonia.FuncUI.Components/Context/Context.Hooks.fs
JordanMarr: I think it needs to live in `Avalonia.FuncUI.Components` because it depends on `useState` and `IWritable<'value>`.
It is designed to only work within component-based UIs.

JaggerJo: Adding a reference to the components project to the Elmish one should be fine for until we move components to the main FuncUI Project. 

This way all the Elmish stuff comes with the Elmish package. 

> It is designed to only work within component-based UIs.

Components will be there by default. Even when using Elmish for the larger part of the app having some state being local is super handy IMHO.

JordanMarr:
* Merged upstream changes to unit test project
* ElmishHook moved into the Elmish project (had to upgrade Elmish.fsproj to .net5.0).
* useElmish `init` now takes a `Cmd` that utilizes `useEffect` hook

src/Avalonia.FuncUI.Elmish/Avalonia.FuncUI.Elmish.fsproj
```diff
@@ -15,15 +15,17 @@
...
 
   <ItemGroup>
+    <Compile Include="ElmishHook.fs" />
     <Compile Include="Library.fs" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Elmish" Version="3.1.0" />
-    <PackageReference Update="FSharp.Core" Version="4.7.2" />
+    <PackageReference Update="FSharp.Core" Version="6.0.1" />
```
sleepyfran, line 24: Not sure if this was automatic, but last time while I was upgrading Avalonia to the latest version you asked me to keep 4.7.2 for better backwards compatibility, so we should probably do the same in this project as well 😃

sleepyfran (approved):

src/Avalonia.FuncUI.Elmish/Avalonia.FuncUI.Elmish.fsproj
JordanMarr: Haha that is true! 

@JaggerJo requested that I make the Elmish project depend on the new Components project. 
The new Components project was net5.0 / fsharp 6.0.1 (and was actually using some new stuff - I think in the form of a Readonly attribute), while all the other projects were netstandard2.0. So the only way to do it was to upgrade the Elmish project.

So a decision needs to be made on whether to stay behind or move ahead to new stuff.
But it looks like maybe that decision has already been made. Thoughts @JaggerJo?

sleepyfran: I'm completely up to upgrade all the FSharp.Core packages to the latest version, so sounds good to me 😃

PR #183, opened 2022-01-10
Keyed test for universal hooks

sleepyfran:
This fixes the issues that were preventing the tests from compile + adds a couple of tests to check the keyed logic inside of the differ.


JaggerJo (approved):
LGTM

PR #179, opened 2021-12-12
React / Sutil inspired state handling

JaggerJo:
This is a prototype for a React/Sutil inspired state handling library for FuncUI. 

There is an example contact app included to test it. 

Screenshot 2021-12-13 at 00 23 49

Feedback is appreciated.

PR #185, opened 2022-01-12
Added AvaloniaUI logo as icon in Examples

houstonhaynes:
There were no window icons in the Examples so I created an "icon.ico" from the AvaloniaUI logo and added it to each sample project. I tried to remain consistent but had some trouble with getting the assets picked up as AvaloniaResources, per se. So I'm importing it directly - which may be a bit naïve - or at least not a canonical example.
AvaloniaUI_logo


sleepyfran:
Awesome, thanks a lot for your contribution 😃 I'll try out later to check if there's any better option to import icons, but if not this should be okay!

If you don't mind waiting, can we delay merging this PR until #179 is merged? There's a lot of examples being added there that could use the icons once it's done as well.

sleepyfran:
@houstonhaynes thanks a lot for the patience! We just got #179 merged, so you'd probably get some merge conflicts on the `.fsproj` files of the samples (they were moved to two different folders). If you could update that and add the same icon to the new projects that were created in the Components folder it'd be great 😃

JordanMarr:
Sorry, didn't mean to close this...was trying to get it merged, but I am bad at git. 🙁

PR #199, opened 2022-03-25
Add README to templates project and point main README to it

sleepyfran:
This PR basically brings the README from the AvaloniaCommunity repo with a couple of updates to include the new Component templates and add a bit of an explanation to distinguish them. Also the Templates section in the main README now points to the project inside of this repo instead of bringing the user to the old one in the AvaloniaCommunity org.

To take a look at how the changes look like you can use:
https://github.com/fsprojects/Avalonia.FuncUI/tree/sleepyfran/chores/template-docs


sleepyfran:
Also @AngelMunoz if you have any time to take a look please do since you brought the templates back to this repo, in case I missed anything 😃

src/Avalonia.FuncUI.Templates/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+dotnet new funcui.full.mvu -n NewApp
+```
+
+## Run the App
+
+All templates in both Components and Elmish flavors should create the same result:
+
+```
+dotnet run NewApp
```
AngelMunoz, line 61: I think we can drop the `NewApp` or otherwise we would need to specify the `--project` flag

```suggestion
dotnet run
```

AngelMunoz (approved):
The changes look great, I just left a comment on the running the app section

JordanMarr (approved):

src/Avalonia.FuncUI.Templates/README.md
sleepyfran: You're absolutely right! Since Jordan already merged this one, I created another one to quickly change it. Thanks! 😄

Issue #203, opened 2022-05-05
Runtime exception when running examples


timwillett4:
The examples are still linking to FSharp.Core 4.7.2 whereas the other projects are using 6.0.2.  This causes a runtime exception when running the examples.

PR #204, opened 2022-05-05
Remove package references to FSharp.Core 4.7.2

timwillett4:
This fixes the following issue:
https://github.com/fsprojects/Avalonia.FuncUI/issues/203

Problem is that some projects explicitly reference FSharp.Core 4.7.2 whereas most projects do not add explicit package reference.  The projects that don't add package reference implicitly add latest version of FSharp.Core which causes the examples to crash with a runtime linker error do to mismatch of FSharp.Core versions.

The approach taking in this PR is simply to remove explicit references to 4.7.2 as to keep all projects consistent.


sleepyfran:
Looks good, just double checked that all the examples work correctly, and all is fine. Thanks a lot for the change!

sleepyfran (approved):

PR #205, opened 2022-05-09
Upgrade to Avalonia 0.10.14

sleepyfran:
Freshly released today, let's keep up with upstream changes:
https://github.com/AvaloniaUI/Avalonia/releases/tag/0.10.14

This also bumps the version of the package to 0.5.1 so that the changes get published.


JordanMarr (approved):
Nice! 👍

PR #207, opened 2022-05-31
string hook identity

JaggerJo:


src/Examples/Component Examples/Examples.ContactBook/Views.fs
```diff
@@ -53,6 +54,42 @@ module CustomHooks =
 
             state
 
+    type IComponentContext with
+
+        member inline ctx.useDeferred<'a>(outer: IWritable<'a>, delay: int, [<CallerLineNumber>] ?lineNumber: int) =
```
JaggerJo, line 59: Needed for creating hooks that contain other hooks. 

Because we identify hooks by the caller line number by default, using the same combined/custom hook in a component leads to issues. This makes it possible to uniquely identify all sub hooks in a combined/custom hook.

JordanMarr:
Solving the hard problems!! 👍

PR #212, opened 2022-07-17
Ensuring unique hook identity for custom hooks (v2)

JaggerJo:
More elegant than what I did in https://github.com/fsprojects/Avalonia.FuncUI/pull/207 .

Basically this allows library authors to create composed hooks without explicitly giving each hook a unique identity.

### DIFF
--- a/src/Examples/Component Examples/Examples.ContactBook/Views.fs
+++ b/src/Examples/Component Examples/Examples.ContactBook/Views.fs
@@ -26,70 +26,70 @@ module CustomHooks =
 
     type IComponentContext with
 
-        member this.useAsync<'signal>(init: Async<'signal>, [<CallerLineNumber>] ?identity: int) : IWritable<Deferred<'signal>> =
-            let identity = Option.get identity
-            let state = this.useState (Deferred.NotStartedYet, true, identity)
+        member ctx.useAsync<'signal>(init: Async<'signal>, [<CallerLineNumber>] ?identity: int) : IWritable<Deferred<'signal>> =
+            ctx.useGrouped (identity.Value, fun ctx ->
+                let state = ctx.useState (Deferred.NotStartedYet, true)
 
-            this.useEffect (
-                handler = (fun _ ->
-                    match state.Current with
-                    | Deferred.NotStartedYet ->
-                        state.Set Deferred.Pending
+                ctx.useEffect (
+                    handler = (fun _ ->
+                        match state.Current with
+                        | Deferred.NotStartedYet ->
+                            state.Set Deferred.Pending
 
-                        Async.StartImmediate (
-                            async {
-                                let! result = Async.Catch init
+                            Async.StartImmediate (
+                                async {
+                                    let! result = Async.Catch init
 
-                                match result with
-                                | Choice1Of2 value -> state.Set (Deferred.Resolved value)
-                                | Choice2Of2 exn -> state.Set (Deferred.Failed exn)
-                            }
-                        )
+                                    match result with
+                                    | Choice1Of2 value -> state.Set (Deferred.Resolved value)
+                                    | Choice2Of2 exn -> state.Set (Deferred.Failed exn)
+                                }
+                            )
 
-                    | _ ->
-                        ()
-                ),
-                triggers = [ EffectTrigger.AfterInit ],
-                callerLineNumber = identity
+                        | _ ->
+                            ()
+                    ),
+                    triggers = [ EffectTrigger.AfterInit ]
+                )
+
+                state
             )
 
-            state
 
     type IComponentContext with
 
         member inline ctx.useDeferred<'a>(outer: IWritable<'a>, delay: int, [<CallerLineNumber>] ?lineNumber: int) =
-            let inner = ctx.useState (outer.Current, identity = $"line: {lineNumber} - inner state")
+            ctx.useGrouped (lineNumber.Value, fun ctx ->
+                let inner = ctx.useState (outer.Current)
 
-            ctx.useEffect (
-                handler = (fun _ ->
-                    inner.Set outer.Current
-                ),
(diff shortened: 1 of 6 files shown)


src/Examples/Component Examples/Examples.ContactBook/Views.fs
```diff
@@ -26,71 +26,71 @@ module CustomHooks =
...
-                                match result with
-                                | Choice1Of2 value -> state.Set (Deferred.Resolved value)
-                                | Choice2Of2 exn -> state.Set (Deferred.Failed exn)
-                            }
-                        )
+                        | _ ->
+                            ()
+                    ),
+                    triggers = [ EffectTrigger.AfterInit ]
+                )
```
JaggerJo, line 53: `ctx.useAsync` with `useGrouped`.

src/Examples/Component Examples/Examples.ContactBook/Views.fs
```diff
@@ -26,71 +26,71 @@ module CustomHooks =
...
 
-            ctx.useEffect (
-                handler = (fun _ ->
-                    timer.Current.Stop()
-                    timer.Current.Start()
-                ),
-                triggers = [ EffectTrigger.AfterChange inner ],
-                identity = $"line: {lineNumber} - inner sync effect"
+                inner
```
JaggerJo, line 91: `ctx.useDeferred` with `useGrouped`

(MISC: useDeferred is a really bad name here. It should be named `debounce` or something)

JaggerJo:
Note that this is a (small) breaking change.

JaggerJo:
@JordanMarr can you take a look?

JordanMarr:
Very nice! I should have some time to look at this in more detail on Wed morning.

JordanMarr:
After looking at your PR, I have some questions.

First, when do I, as an end user, need to be cognizant of line numbers? Is it anytime I am creating a hook, or does this only apply to "composed hooks"? 

So then let's say I make a `useCounter` hook that encapsulates a `useState` hook; is that a composed hook?
Note: this example is still based off of the master branch. Also, I added the `?identity` arg only because I copied and pasted from another hook. Should I have used it somewhere in my hook? Does your PR make it unnecessary?

```F#
module Examples.CounterWithCustomHook

open Avalonia.Layout
open Avalonia.Controls
open Avalonia.FuncUI
open Avalonia.FuncUI.DSL
open Avalonia.FuncUI.Types

[<AutoOpen>]
module CustomHook = 
    open System.Runtime.CompilerServices

    type IComponentContext with
        member ctx.useCounter(initialCount: int, [<CallerLineNumber>] ?identity: int) =
            let counter = ctx.useState(initialCount)
            let increment () = counter.Set(counter.Current + 1)
            let decrement () = counter.Set(counter.Current - 1)
            counter.Current, increment, decrement

let cmp () = Component (fun ctx ->
    let count, increment, decrement = ctx.useCounter 0
    
    Grid.create [
        Grid.rowDefinitions "20"
        Grid.columnDefinitions "*, 80, *"
        Grid.children [

            StackPanel.create [
                Grid.column 1
                Grid.row 1

                StackPanel.children [
                    // Increment
                    Button.create [
                        Button.content "▲"
                        Button.horizontalAlignment HorizontalAlignment.Stretch
                        Button.horizontalContentAlignment HorizontalAlignment.Center
                        Button.onClick (fun _ -> increment ())
                    ]

                    // Current count
                    TextBlock.create [
                        TextBlock.text (string count)
                        TextBlock.horizontalAlignment HorizontalAlignment.Center
                    ]

                    // Decrement
                    Button.create [
                        Button.content "▼"
                        Button.horizontalAlignment HorizontalAlignment.Stretch
                        Button.horizontalContentAlignment HorizontalAlignment.Center
                        Button.onClick (fun _ -> decrement ())
                    ]

                ]
            ]

        ]
    ]
    :> IView
)

```

JaggerJo:
@JordanMarr not sure if the additional complexity of using the caller line number is worth it vs. using the calling order (like react). We might want to consider switching to the react way.

sleepyfran:
> @JordanMarr not sure if the additional complexity of using the caller line number is worth it vs. using the calling order (like react). We might want to consider switching to the react way.

I honestly do think conditional hooks are a useful thing, I'm doing React most of my dev day and there's a bunch of situations when you want a hook with side effects to be executed under a certain condition and having to go through all the trouble of setting up a new component just to be able to conditionally call a hook is a hassle. I haven't seen the code of this PR yet, but unless it's an incredibly complex logic to allow this or it affects performance greatly I'd say it's a good thing to have.

JordanMarr:
Sorry for the delay, it's been a busy couple of weeks.

That's definitely an improvement!

My only concern is that the concept of passing a line number at all is hook internal implementation details leaking. If line numbers are confusing to me, they will be confusing to others too. 

Is it fair to say that the tradeoff is: 
_no conditional hooks and no line numbers_ vs. _conditional hooks with line numbers_?

If so, conditional hooks would require a lot less knowledge -- only one simple rule for end users to remember: "no conditional  hooks!"

JordanMarr (approved):
Approved (with comment that it will be confusing to users)

JaggerJo:
> Approved (with comment that it will be confusing to users)

I agree with @JordanMarr and think we should move to simpler `calling order` based identity. Also think that conditional hooks are sometimes pretty useful @sleepyfran but we could still allow explicit hook identity to enable this to more experienced users if we really wanted.

I'm on vacation starting from today, but think the "a hooks identity is determined by the caller order" is pretty simple to implement - and what we should do IMO.

JordanMarr:
That sounds like a best of both worlds approach. 
The common scenario “just works” if you follow React rules and avoid conditional hooks, but a line number can be passed for more niche, advanced cases.

JaggerJo:
Closed in favour of #214

PR #214, opened 2022-08-21
Simplify Hook Identity

JaggerJo:
This PR simplifies how StateHooks & EffectHooks are identified. Previously there was some "Magic" ([CallerLineNumber](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerlinenumberattribute?view=net-6.0)) going on so that hooks automatically got a unique identity that was independent from the calling order.

There were some issues when hooks were called outside of the render function (for example in a custom hook that composes some basic hooks). While it was/is possible to work around those issues it makes the whole model a lot more complicated than required for most cases.

The simpler implementation basically copies react and uses the calling order as the identity.

⚠️ This makes it illegal to have a conditional `useState`/`useEffect`!

This should not break existing applications too much as long as they don't:
- correctly implement custom hooks (simply solved by getting rid of the passed identity)
- use hooks conditionally (requires work and might crash the app)

@sleepyfran I know you would prefer having conditional hooks, but I think it's too complicated to understand with all carvets for most users. If we find a way to add this back without making authoring custom hooks more complicated I'm totally in for it.


JaggerJo:
This worked first try - I clicked through the example apps, but I'm still suspicious 😁

PR #218, opened 2022-09-08
Add onSelectedItemChanged to TreeView

Numpsy:
Hi,
I was having a go at a UI using a TreeView and I needed to handle a selection changed event, and I didn't see it in the current version.
I added a local extension method and that seemed to work ok, but I thought i'd see if it could be added to the build in bindings (I'm not sure what the situation is with just adding extra bindings to existing controls?)

Thanks


JaggerJo:
LGTM, just one thing:

I've seen you created it based on the `TreeView.SelectedItemProperty` property. Is there no RoutedEvent that could be used for this? 

Thanks for the PR!

Numpsy:
I only see a regular EventHander (https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/TreeView.cs#L72), though I am working out how all this works as I go so I might be missing something somewhere

JaggerJo:
LGTM

PR #220, opened 2022-10-08
Add support for RichTextBlock

sleepyfran:
This PR attempts to add support for the new RichTextBlock introduced in Avalonia v11. Although adding the new control is pretty straightforward, the main challenge comes from the new inline elements that were introduced for the new control, since these should be used the same way as controls are used but are not controls per se.

## How the new elements work
The new control contains a property with a collection of [Inlines](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Inline.cs), which are themselves subclasses of StyledElements. Each inline then defines how that portion of the text should be styled, with `Run` being a portion of text and `Span` being a group for other inline elements. All these are then collected by the control and rendered.

## Implementation for FuncUI
For a comprehensive demo of all the elements take a look at the new example I created 😄 

## Demo time!
Inline

## To Do
- [x] Add styling support for [Run](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Run.cs)
- [x] Add bindings for [Span](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Span.cs)
- [x] Add bindings for [Bold](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Bold.cs)
- [x] Add bindings for [Italic](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Italic.cs)
- [x] Add bindings for [Underline](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/Underline.cs)
- [x] Add bindings for [LineBreak](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Documents/LineBreak.cs)
- [x] Validate that the patcher changes don't break anything (so far I've just run a couple of examples, but I'd like to validate the rest)
- [ ] Check how can we make inline elements be `IView<Inline>` instead of the non-generic `IView` while preventing unnecessary castings

### DIFF
--- a/src/Avalonia.FuncUI/VirtualDom/VirtualDom.fs
+++ b/src/Avalonia.FuncUI/VirtualDom/VirtualDom.fs
@@ -10,7 +10,7 @@ module rec VirtualDom =
     let create (view: IView) : IControl =
         view
         |> ViewDelta.From
-        |> Patcher.create :?> IControl (* TODO: Can we remove this cast? *)
+        |> Patcher.create :?> IControl
 
     let update (root: IControl, last: IView, next: IView) : unit =
         let delta = Differ.diff(last, next)
@@ -76,11 +76,11 @@ module rec VirtualDom =
             | ValueSome delta ->
                 match control.GetType () = delta.ViewType && not delta.KeyDidChange with
                 | true -> Patcher.patch (control, delta)
-                | false -> host.Child <- (Patcher.create delta) :?> IControl (* TODO: Can we remove this cast? *)
+                | false -> host.Child <- (Patcher.create delta) :?> IControl
             | ValueNone ->
                 host.Child <- null
 
         | ValueNone ->
             match delta with
-            | ValueSome delta -> host.Child <- (Patcher.create delta) :?> IControl (* TODO: Can we remove this cast? *)
+            | ValueSome delta -> host.Child <- (Patcher.create delta) :?> IControl
             | ValueNone -> host.Child <- null
--- a/src/Avalonia.FuncUI/VirtualDom/VirtualDom.fs
+++ b/src/Avalonia.FuncUI/VirtualDom/VirtualDom.fs
@@ -10,7 +10,7 @@ module rec VirtualDom =
     let create (view: IView) : IControl =
         view
         |> ViewDelta.From
-        |> Patcher.create
+        |> Patcher.create :?> IControl (* TODO: Can we remove this cast? *)
 
     let update (root: IControl, last: IView, next: IView) : unit =
         let delta = Differ.diff(last, next)
@@ -76,11 +76,11 @@ module rec VirtualDom =
             | ValueSome delta ->
                 match control.GetType () = delta.ViewType && not delta.KeyDidChange with
                 | true -> Patcher.patch (control, delta)
-                | false -> host.Child <- Patcher.create delta
+                | false -> host.Child <- (Patcher.create delta) :?> IControl (* TODO: Can we remove this cast? *)
             | ValueNone ->
                 host.Child <- null
 
         | ValueNone ->
             match delta with
-            | ValueSome delta -> host.Child <- Patcher.create delta
+            | ValueSome delta -> host.Child <- (Patcher.create delta) :?> IControl (* TODO: Can we remove this cast? *)
             | ValueNone -> host.Child <- null
--- a/src/Avalonia.FuncUI.UnitTests/VirtualDom/VirtualDom.PatcherTests.fs
+++ b/src/Avalonia.FuncUI.UnitTests/VirtualDom/VirtualDom.PatcherTests.fs
@@ -44,19 +44,19 @@ module PatcherTests =
 
     [<Fact>]
     let ``Patch Styles, Classes or Resources`` () =
-        let stylesGetter: IControl -> obj = (fun c -> (c :?> StyledElement).Styles :> obj)
-        let stylesSetter: IControl * obj -> unit =
+        let stylesGetter: IAvaloniaObject -> obj = (fun c -> (c :?> StyledElement).Styles :> obj)
+        let stylesSetter: IAvaloniaObject * obj -> unit =
             (fun (c, v) ->
                 let se = (c :?> StyledElement)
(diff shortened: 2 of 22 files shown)


JaggerJo:
Hey @sleepyfran, nice work! 

I've been waiting for more advanced text formatting support in Avalonia for a while now 😁 nice that it's finally here. 

-> Yes, thing we need to use `IAvaloniaObject` everywhere in the differ/patcher. 

I think it would be possible to treat `Inlines` as a kind of `Content` / expand what we see as content.  

https://github.com/fsprojects/Avalonia.FuncUI/blob/3a32ee2bc887118878e4be85cb8a0a1dd41044dc/src/Avalonia.FuncUI/Types.fs#L89-L92 

https://github.com/fsprojects/Avalonia.FuncUI/blob/3a32ee2bc887118878e4be85cb8a0a1dd41044dc/src/Avalonia.FuncUI/Types.fs#L53-L59

https://github.com/fsprojects/Avalonia.FuncUI/blob/3a32ee2bc887118878e4be85cb8a0a1dd41044dc/src/Avalonia.FuncUI/Types.fs#L126-L135

JaggerJo:
(`IView` might be a bit misleading then, but we could consider renaming it and creating an alias for compatibility)

It should be possible to only allow certain elements as inlines by using a `IView<Inline> list`

Could look something like this:
```fsharp

    type RichTextBlock with
            
        static member inlines<'t when 't :> RichTextBlock>(value: IView<Inline> list) : IAttr<'t> =
            let getter : ('t -> obj) = (fun control -> control.Inlines :> obj)
             
            AttrBuilder<'t>.CreateContentMultiple("Inlines", ValueSome getter, ValueNone, value)

```

sleepyfran:
Thanks @JaggerJo! That's a really good idea, I'm not sure why I went the way of creating a new category when expanding the content would fit nicely, I did so many blind changes in the beginning trying to make it work that I somehow missed it 😅 

Will try to expand the content to accept inline elements, should be pretty straightforward and most of the types I've defined so far should be reusable 😄

sleepyfran:
@JaggerJo please take a look at the last commit 😄  I've pretty much switched to `IAvaloniaObject` everywhere it was possible and switched to `IView` for the inline elements. There's just a few _buts_ I've found on the way, I've annotated them with TODOs inside the code, but to have them easily listed:
- Because `Inline` is an abstract class we can't return it as the generic type of `IView<>` in the inline elements, since otherwise the activator won't be able to create the instance later in the patcher. This makes it impossible (at least that I've found) to use `IView<Inline>` in the actual `RichTextBlock`since we can't do the cast to `IView<Inline>` later on. I've made it `IView list` for now, but it'd be great to narrow it to only inline elements.
- I haven't found a way of getting rid of a couple `IControl` references, namely subscriptions have a hard requirement of `IInteractive` and `IAvaloniaObject` that is only found on controls. I've just done a pattern match on the casting and fail if we're attempting to subscribe to a non-control, which I believe should be fine.
- There's a couple of ugly casts to `IControl` in the `VirtualDom` root that I'm not sure if we can somehow remove. I'm pretty sure I'm missing something here, so if you see a way please let me know!

Other than that, inlines are working as ContentMultiple, so awesome job making the initial architecture so damn extensible!

src/Avalonia.FuncUI/VirtualDom/VirtualDom.fs
```diff
@@ -76,11 +76,11 @@ module rec VirtualDom =
             | ValueSome delta ->
                 match control.GetType () = delta.ViewType && not delta.KeyDidChange with
                 | true -> Patcher.patch (control, delta)
-                | false -> host.Child <- Patcher.create delta
+                | false -> host.Child <- (Patcher.create delta) :?> IControl (* TODO: Can we remove this cast? *)
```
JaggerJo, line 79: `IControl.Content` is expected to be an `IControl` so the cast is good IMHO.

JaggerJo:
> @JaggerJo please take a look at the last two commits 😄 I've pretty much switched to `IAvaloniaObject` everywhere it was possible and switched to `IView` for the inline elements. There's just a few _buts_ I've found on the way, I've annotated them with TODOs inside the code, but to have them easily listed:

> * Because `Inline` is an abstract class we can't return it as the generic type of `IView<>` in the inline elements, since otherwise the activator won't be able to create the instance later in the patcher. This makes it impossible (at least that I've found) to use `IView<Inline>` in the actual `RichTextBlock`since we can't do the cast to `IView<Inline>` later on. I've made it `IView list` for now, but it'd be great to narrow it to only inline elements.

Hmm, yeah - need to take a deeper look. Would be really nice if we could constrain the type of View we support here (and in. a few other places). 

> * I haven't found a way of getting rid of a couple `IControl` references, namely subscriptions have a hard requirement of `IInteractive` and `IAvaloniaObject` that is only found on controls. I've just done a pattern match on the casting and fail if we're attempting to subscribe to a non-control, which I believe should be fine.

agree 👍

> * There's a couple of ugly casts to `IControl` in the `VirtualDom` root that I'm not sure if we can somehow remove. I'm pretty sure I'm missing something here, so if you see a way please let me know!

Think this is fine. In some cases we really expect a Control, not an AvaloniaObject. 

> Other than that, inlines are working as ContentMultiple, so awesome job making the initial architecture so damn extensible!

Thanks ☺️

sleepyfran:
@JaggerJo with the latest commits all the elements are implemented and all of them stylable 😃  I've made a utility function for Bold, Italic, Underline and LineBreak so that they don't require the whole ordeal to set-up a simple text line, but I'm not so sure about the name `simple`. If you have any suggestions here let me know!

So with this I think the only thing left is to check if we could somehow remove the `IView` requirement, but not sure if we should do this here or try to address it in another PR (iirc there was already a ticket/PR for unifying both of them)

JaggerJo (approved):
@sleepyfran Great work, let's merge! 🎉

PR #224, opened 2022-10-21
Add `simple` function to Run elements

sleepyfran:
Super simple change that somehow I forgot to add when in #220. This follows the same convention as Bold, LineBreak, etc.


JaggerJo:
Maybe we should name it `Run.createText: string -> View<Run>`?

JaggerJo (approved):
LGTM

PR #225, opened 2022-10-25
Fix MaskedTextBox attributes

sleepyfran:
The control bindings introduced in #217 do not accept `MaskedTextBox` attributes currently because the attrs parameter is set to `TextBox`. Changing to `MaskedTextBox` accepts both, since the later already inherits from `TextBox` anyway.

Edited the contacts example to display this masked text box with a mask:

image

_(Won't be uploading this since some of the phone numbers don't follow that format and they don't work correctly, but it shows that the bindings work as the `mask` property is only available on `MaskedTextBox` and it was not being accepted before)_

Also bumping the preview version since I need this on one of my projects 😃


JaggerJo (approved):
LGTM

Issue #227, opened 2022-11-01
Flyout component support


picolino:
According to docs (https://docs.avaloniaui.net/docs/controls/flyouts) Avalonia has Flyout components, but in Avalonia.FuncUI no implementation for that component, so I tried to add it by myself and I've got a problem that I'm stuck with.

I've implemented basic frame for new component:

```fs
namespace Avalonia.FuncUI.DSL

[<AutoOpen>]
module Flyout =
    open Avalonia.Controls
    open Avalonia.FuncUI.Types
    open Avalonia.FuncUI.Builder
    open Avalonia.Controls.Primitives
    
    let create (attrs: IAttr<Flyout> list): IView<Flyout> =
        ViewBuilder.Create<Flyout>(attrs)

    type FlyoutBase with

        /// <summary>
        /// A value indicating whether the flyout is visible.
        /// </summary>
        static member isOpen<'t when 't :> FlyoutBase>(value: bool) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<bool>(FlyoutBase.IsOpenProperty, value, ValueNone)
            
        /// <summary>
        /// A value indicating how the flyout is positioned.
        /// </summary>
        static member placement<'t when 't :> FlyoutBase>(value: FlyoutPlacementMode) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<FlyoutPlacementMode>(Flyout.PlacementProperty, value, ValueNone)
            
        /// <summary>
        /// A value indicating flyout show mode.
        /// </summary>
        static member showMode<'t when 't :> FlyoutBase>(value: FlyoutShowMode) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<FlyoutShowMode>(Flyout.ShowModeProperty, value, ValueNone)
            
    type Flyout with

        static member content<'t when 't :> Flyout>(value: IView option) : IAttr<'t> =
            AttrBuilder<'t>.CreateContentSingle(Flyout.ContentProperty, value)

        static member content<'t when 't :> Flyout>(value: IView) : IAttr<'t> =
            value
            |> Some
            |> Flyout.content
            
    type Button with
    
        /// <summary>
        /// A value for flyout view placement for button.
        /// </summary>
        static member flyout<'t when 't :> Button>(value: IView option) : IAttr<'t> =
            AttrBuilder<'t>.CreateContentSingle(Button.FlyoutProperty, value)
    
        /// <summary>
[...]

PR #229, opened 2022-11-03
Flyout component support

picolino:
Add Flyout component support with demo according to issue #227

### DIFF
--- a/src/Avalonia.FuncUI/DSL/Flyout.fs
+++ b/src/Avalonia.FuncUI/DSL/Flyout.fs
@@ -1,34 +1,32 @@
 ﻿namespace Avalonia.FuncUI.DSL
 
+open Avalonia.Controls.Templates
+
 [<AutoOpen>]
 module Flyout =
     open Avalonia.Controls
     open Avalonia.FuncUI.Types
     open Avalonia.FuncUI.Builder
     open Avalonia.Controls.Primitives
-    
+
     let create (attrs: IAttr<Flyout> list): IView<Flyout> =
         ViewBuilder.Create<Flyout>(attrs)
 
     type FlyoutBase with
-            
+
         /// <summary>
         /// A value indicating how the flyout is positioned.
         /// </summary>
         static member placement<'t when 't :> FlyoutBase>(value: FlyoutPlacementMode) : IAttr<'t> =
             AttrBuilder<'t>.CreateProperty<FlyoutPlacementMode>(FlyoutBase.PlacementProperty, value, ValueNone)
-            
+
         /// <summary>
         /// A value indicating flyout show mode.
         /// </summary>
         static member showMode<'t when 't :> FlyoutBase>(value: FlyoutShowMode) : IAttr<'t> =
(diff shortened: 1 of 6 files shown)


src/Avalonia.FuncUI/DSL/Flyout.fs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        static member flyoutPresenterClasses<'t when 't :> Flyout>(value: string list) : IAttr<'t> =
+            let getter : ('t -> string list) = (fun control -> control.FlyoutPresenterClasses |> Seq.map id |> List.ofSeq)
+            let setter : ('t * string list -> unit) = (fun (control, value) ->
+                control.FlyoutPresenterClasses.Clear()
+                control.FlyoutPresenterClasses.AddRange(value))
+            
+            AttrBuilder<'t>.CreateProperty<string list>("FlyoutPresenterClasses", value, ValueSome getter, ValueSome setter, ValueNone)
+            
+    type Button with
```
picolino, line 51: Not sure if I did this correctly, probably I should move extension methods for `Button` in `Button.fs` file instead, pls note me if so.

sleepyfran: Yeah, it'd be better to have this on the `Button` file with the rest of the bindings for it 😃

src/Avalonia.FuncUI/DSL/Flyout.fs
```diff
@@ -0,0 +1,65 @@
...
+        static member placement<'t when 't :> FlyoutBase>(value: FlyoutPlacementMode) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<FlyoutPlacementMode>(FlyoutBase.PlacementProperty, value, ValueNone)
+            
+        /// <summary>
+        /// A value indicating flyout show mode.
+        /// </summary>
+        static member showMode<'t when 't :> FlyoutBase>(value: FlyoutShowMode) : IAttr<'t> =
+            AttrBuilder<'t>.CreateProperty<FlyoutShowMode>(FlyoutBase.ShowModeProperty, value, ValueNone)
+            
+    type Flyout with
```
sleepyfran, line 27: I see in the documentation that there's also a `MenuFlyout`, could we also support this in here? I haven't checked in detail, if it'd be too much work we can address it separately but just in case it's just a few more lines.

src/Avalonia.FuncUI/DSL/Flyout.fs
sleepyfran: Yeah, I think it'd be better to have it in the Button file 😃

picolino: Done

src/Avalonia.FuncUI/DSL/Flyout.fs
picolino: Done, I placed it inside `Flyout.fs` file, not sure again if that corrects, note me if you want it to be in separated file.

sleepyfran (approved):
Looks good, awesome job!

Issue #232, opened 2022-11-12
11.0.0-preview4 vs. RichTextBlock


Numpsy:
I just tried to update an application that was using Avalonia 11.0.0-preview3 to the new 11.0.0-preview4 build and got a bunch of errors from FuncUI about ```RichTextBlock``` not existing.

I had a look at the Avalonia change history and saw https://github.com/AvaloniaUI/Avalonia/pull/9245 which seems to have renamed the control

JaggerJo:
@Numpsy interesting 🤨

Would you mind creating a PR that aligns naming?

Otherwise it might take a bit until it's fixed.

Numpsy:
I'll have a look later

Numpsy:
I started to have a look, and then fell over another breaking Avalonia change in #233 

Another related thing as well - version 11 of Avalonias Blazor integration has been retargetted against .NET 7.0, so I think the web assembly playground app will need changing to match

sleepyfran:
Closed in #234

PR #233, opened 2022-11-12
Remove the header properties from DatePicker and TimePicker

Numpsy:
I was looking at #232 and then also fell over https://github.com/AvaloniaUI/Avalonia/issues/9136, for which the header properties have been removed from the DatePicker and TimePicker controls.

So - this removes those.

That also required changes to the control catalog (I changed it to put a text block above the sample date/time controls to contain the header text)


JaggerJo (approved):
LGTM

PR #237, opened 2022-12-26
Transform comparison

Thecentury:
Provided a custom equality comparer for types implementing the `ITransform` interface. 
Types implementing the `ITransform` don't override equality methods, so using them in FuncUI means reassignments on every view generation (if transforms are created on the fly instead of reusing the same instance, producing an object not equal to the previous one even if it's created with the same constructor arguments as default `Equals` compares references). And such reassignment means an additional layout/redraw pass.


JordanMarr (approved):
Very nice optimization!

PR #239, opened 2022-12-26
Fix compilation warnings

Thecentury:
- Removed unused namespace imports
- Prefixed unused parameters with an underscore
- Match all possible `Orientation` values
- Override the `GetHashCode` method of the `ViewDelta` type

The `Avalonia.FuncUI.Diagnostics.Reflector` type mentions a `Maecos.Shared.NonEmptyList` type (note a possible type in "Maecos") — https://github.com/fsprojects/Avalonia.FuncUI/blob/master/src/Avalonia.FuncUI.Diagnostics/Lib/Lib.Reflector.fs#L46. Is it a correct string or a different type was meant there?


Numpsy:
It would be nice to get the existing warnings fixed (there are a few new ones with the latest Avalonia builds due to more things being obsoleted, and the noise makes them harder to see)

JordanMarr (approved):

Issue #222, opened 2022-10-19
Avalonia.FuncUI v1.0 release


sleepyfran:
Opening this issue to track the work needed before we release the non-pre-release version of v0.6

# Before release
- [x] Wait for Avalonia v11 to be released (duh!)
- [x] Adapt to "The Great Interface Removal" introduced in https://github.com/AvaloniaUI/Avalonia/pull/9553 (Addressed in #246)
- [x] Remove mention to old docs site in README and point all of them to the [docs site under Avalonia's domain](https://funcui.avaloniaui.net/)
- [ ] Create a Migration Guide in the docs
  - [ ] Interface removals (#246)
  - [ ] Changes for Elmish v4 (#248)
  - [ ] Package name changes (#242)
- [ ] Migrate templates to new version (includes adding dependency of the newly separated `Avalonia.Themes.Fluent`)
- [x] Deprecate Avalonia.FuncUI.DSL package from Nuget. See [here](https://learn.microsoft.com/en-us/nuget/nuget-org/deprecate-packages). I believe this can only be done by you @JaggerJo?

Since 4564a6d98c55e3b9e0437e4d8aecfb41bfed9b5b the DSL project lives inside of the main FuncUI package and upgrading from a previous version that has the a package reference to the DSL makes all this funky errors where half of the imports seem unused and all the controls stop working:
<img width="564" alt="image" src="https://user-images.githubusercontent.com/6024783/196653338-242150c9-94e1-4a6c-aff2-bf8c5fd1856d.png">

The fix is, of course, to remove the package reference to the DSL but I think we should make this clearer to the user.

# After release
- [ ] Create documentation for new inline components
- [ ] Party? 🎉

JaggerJo:
Like the `Party 🎉` part of it 😁

JaggerJo:
I can deprecate the DSL package shortly before we plan to release the new version 👍

JaggerJo:
I'm not happy at all with the documentation situation right now. 🙃

sleepyfran:
> I'm not happy at all with the documentation situation right now. 🙃

Yeah, actually I just realized that we have the GitBook under the Avalonia's domain up and working, I'll add another task here to sunset the current one we have (https://avaloniacommunity.github.io/Avalonia.FuncUI.Docs/) and add the one under Avalonia's domain to the README. What else do you think we should do to improve it? I'm all up for making the docs better, it's true that we're lacking a bit here.

[...]

PR #244, opened 2023-01-14
update funcui to funcUI

ShalokShalom:
this adds a correction to the spelling issue introduced with the new templates


ShalokShalom:
Also: I am confused, about the documentation of this project including the installation of the templates, and not of the libary itself. 

Why that is?

sleepyfran:
Oops, thanks a lot for addressing this! Just double confirmed with `dotnet new --list` and we have indeed uppercase in the templates. Thank you for the fix 😄 

About your question: it's something we're planning to update in the docs once we update them, you can see more details in #222

sleepyfran (approved):

Issue #240, opened 2023-01-05
Avalonia.FuncUI.Elmish is incompatible with Elmish 4


vincentremond:
The method `withSyncDispatch` seems to have been removed in Elmish 4.
But the package `Avalonia.FuncUI.Elmish` [seems to rely on it](https://github.com/fsprojects/Avalonia.FuncUI/blob/master/src/Avalonia.FuncUI.Elmish/Library.fs#L31).

At runtime I get this error : 
```
Unhandled exception. System.MissingMethodException: Method not found: 'Elmish.Program`4<System.__Canon,System.__Canon,System.__Canon,System.__Canon> Elmish.ProgramModule.withSyncDispatch(Microsoft.FSharp.Core.FSharpFunc`2<Microsoft.FSharp.Core.FSharpFunc`2<System.__Canon,Microsoft.FSharp.Core.Unit>,Microsoft.FSharp.Core.FSharpFunc`2<System.__Canon,Microsoft.FSharp.Core.Unit>>, Elmish.Program`4<System.__Canon,System.__Canon,System.__Canon,System.__Canon>)'.
   at Avalonia.FuncUI.Elmish.Program.withHost[arg,model,msg,a](IViewHost host, Program`4 program)
   at HueModoro.Shell.MainWindow..ctor() in D:\XXX\Shell.fs:line 86
   at HueModoro.App.OnFrameworkInitializationCompleted() in D:\XXX\Program.fs:line 21
   at Avalonia.Controls.AppBuilderBase`1.Setup() in /_/src/Avalonia.Controls/AppBuilderBase.cs:line 312
   at Avalonia.Controls.AppBuilderBase`1.SetupWithLifetime(IApplicationLifetime lifetime) in /_/src/Avalonia.Controls/AppBuilderBase.cs:line 179
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime[T](T builder, String[] args, ShutdownMode shutdownMode) in /_/src/Avalonia.Controls/ApplicationLifetimes/ClassicDesktopStyleApplicationLifetime.cs:line 208
   at HueModoro.Program.main(String[] args) in D:\XXX\Program.fs:line 28
   ```

JordanMarr:
Thanks for posting the issue. 
I am looking into this as well as other Elmish 4 changes today.

JordanMarr:
The latest prerelease (available now) has been upgraded to Elmish v4. 

[...]

PR #245, opened 2023-01-15
Strictify Elmish dependency version

NicoVIII:
Elmish 4.0 introduced breaking changes, therefore Avalonia.FuncUI.Elmish doesn't work with that version (see #240). To avoid a non-working package resolution, I strictified the version requirement.

After that the generated .nupkg had the following dependencies:
```
Dependencies:
  net6.0:
    JaggerJo.Avalonia.FuncUI: '>= 0.6.04'
    Elmish: '>= 3.1.0 && < 4.0.0'
    FSharp.Core: '>= 7.0.0'
```


JordanMarr (approved):
Looks good, thanks!

PR #246, opened 2023-01-15
[WIP] Update to build with the latest Avalonia 11 changes

Numpsy:
A WIP change, to try to make it build against the latest Avalonia 11 nightly build.

Most of the changes are to replace uses of interfaces with their base classes to make it work with the Avalonia changes from https://github.com/AvaloniaUI/Avalonia/pull/9553

Note: I've had the control catalog working with thest changes, *except* that horizontally resizing the window corrupts some of the contents. That may or may not be down to the issue described at https://github.com/AvaloniaUI/Avalonia/issues/9975


src/Avalonia.FuncUI/Library.fs
```diff
@@ -33,7 +33,7 @@ module internal Extensions =
         member this.SubscribeWeakly(callback: 'a -> unit, target) =
             Observable.subscribeWeakly(this, callback, target)
 
-    type IInteractive with
+    type Interactive with
```
Numpsy, line 36: Another change that effects this bit is that the Avalonia change at https://github.com/AvaloniaUI/Avalonia/pull/9749 has removed its usage of System.Reactive and replaced it with a bunch of (mostly internal) things.

I added a local reference to System.Reactive to make it build, but i'm not sure of the 'best' approach (still learing how all this hangs together) - e.g. Avalonia now seems to have an internal implementation of Observable.Create that uses a minimal local class, so I wondered if doing something like

```fsharp
{ new IObservable<'args>
  with member this.Subscribe(observer: IObserver<'args>) = sub.Invoke(observer) }
```

would work instead of importing Reactive just for Observable.Create?

There are also 3 references to Observable.Subscribe that require Reactive, but the comment at https://github.com/AvaloniaUI/Avalonia/pull/9749#issuecomment-1378290972 suggests there might be more changes coming on the Avalonia side there

Numpsy:
There is also a break in the web assembly playground project due to Avalonia.Web.Blazor having been renamed to Avalonia.Browser.Blazor, but that doesn't doesn't seem to currently be attached to the solution?

JaggerJo:
> There is also a break in the web assembly playground project due to Avalonia.Web.Blazor having been renamed to Avalonia.Browser.Blazor, but that doesn't doesn't seem to currently be attached to the solution?

No, The WASM Example even had runtime errors last time I tried it with v11.

src/Avalonia.FuncUI/Library.fs
JaggerJo: should be good 👍

Numpsy:
Ok, I've updated the project anyway, something else to test with the new versions

src/Avalonia.FuncUI/Library.fs
Numpsy: The general bit about System.Reactive, or the local object expression suggestion?

JaggerJo: I'd prefer not having a reference to 'System.Reactive'. So the object expression seems good to me.

Numpsy:
Another breaking change has appeared in the Avalonia 11 nightlys - the change at https://github.com/AvaloniaUI/Avalonia/pull/9677 has changed some virtualization related things, which causes a number if build breaks due to properties having been removed or renamed

src/Avalonia.FuncUI/Library.fs
Numpsy: Would there be any value in breaking bits like this into separate PRs (in cases where they can be done without updating the Avalonia version)?

JordanMarr: The two cases where breaking this PR into smaller bits might make sense:
1) if the change could be released as a new preview version that would provide immediate value for end users.
2) if the change is big enough (touches enough files) that it will become harder to merge the longer it sits.

Numpsy: The question was mainly that you don't have to do it all in one go (as you could leave the FuncUI code using a local System.Reactive reference to begin with and fix it later, *or* you could try removing local dependencies on System.Reactive without updating Avalonia, e.g. #261 ).
The value is probably more that it'd be more self contained to test.

Numpsy:
I was holding off on trying to make a couple of other System.Reactive uses (```Subscribe``` calls using callbacks and cancellation tokens) self contained in case other Avalonia changes avoid the need (e.g. https://github.com/AvaloniaUI/Avalonia/pull/10105), but looking at what Reactive actually does I wonder if something using the built in ```Observable.subscribe``` along the lines of 

```fsharp
type IObservable<'a> with
        member this.Subscribe (callback: 'a -> unit, token: CancellationToken) =
            let disposable = Observable.subscribe callback this
            token.Register(fun () -> disposable.Dispose())
```

might work?

Numpsy:
Updated with a couple more breaking API changes in Avalonia (this time just in the sample apps though)

https://github.com/AvaloniaUI/Avalonia/pull/8166
https://github.com/AvaloniaUI/Avalonia/pull/10141

Numpsy:
Remove the ItemsRepeater bindings after https://github.com/AvaloniaUI/Avalonia/pull/10112

Numpsy:
Ok, updated to Avalonia 11 preview 5.

Seems to be working ok, but still has a System.Reactive reference to get a couple of uses of ```Subscribe``` (see previous Q about if a simple local extension method will work instead)

JaggerJo:
Wonder when all of this great stuff will land in a new public avalonia preview 😀

Numpsy:
> Wonder when all of this great stuff will land in a new public avalonia preview 😀

Looks like it just did: 
image

JaggerJo:
Nice! Looking forward to merging your changes and also releasing a preview version.

Numpsy:
hmm, forgot about the web assembly playground project, might need more tweaks there (although, it could maybe also be simplified to use the new self-hosted WASM support rather than needing Blazor?)

JaggerJo:
Whatever works 😉

Numpsy:
I've updated the playground project so it builds with Preview 5, but I got a javascript error when running it - so that bit needs another look later

JaggerJo:
I doesn't work currently, so we can definitely fix it after merging. 👍

JaggerJo:
Merged! Great work 🔥

PR #247, opened 2023-01-15
Update the style references in the Examples.Presso project

Numpsy:
I was getting a build error about this with the latest Avalonia 11, but it appears that there is already a runtime error with the current code (were the xaml files copied from Example.Elmish.Presso without changing the namespaces?)


JaggerJo (approved):
Yes, strange that it was unnoticed.

PR #248, opened 2023-01-15
Elmish v4 Changes

JordanMarr:
# Overview of Elmish v4 Changes

Elmish changes are listed [here](https://github.com/elmish/elmish/blob/d3e6272eec942d81ef77026bce7594be1d0246a4/RELEASE_NOTES.md) in the Elmish repo.

## 1) Breaking Changes to UI Thread Sync
* `Program.withSyncDispatch` has been removed in Elmish v4 because the platform specific `Program.runWith___` functions should instead be provided by the library authors.

**Additional info** 
the purpose of the "sync dispatch" is to ensure that UI changes coming from non-UI threads can be properly synchronized via the underlying platform specific dispatcher to avoid race conditions. 
Previously, FuncUI _automatically_ configured this within the `Program.withHost` function, on behalf of the user, to guard against race conditions that could occur if they called `dispatch` from an `async` handler.

**Avalonia.FuncUI upgrade path**
A new `Program.runWithAvaloniaSyncDispatch` function has been added that conforms to the new expectation that this should be handled by the library. Users that are calling `dispatch` from within `async` blocks will need to know to use this instead of `Program.run`. If they don't, the will get exceptions, which will then force them to use this new function.

Unfortunately, this does mean that users will need to be responsible for using `Program.runWithAvaloniaSyncDispatch` vs `Program.run`. However, we can accomplish this via a "breaking changes update guide". Also, I think it is common for users to have some understanding of thread dispatching concerns and how to handle it properly. 

**Old Elmish v3 Sample with async blocks**
```F#
Program.mkProgram init update view
|> Program.withHost this
|> Program.run
```
**New Elmish v4 Sample with async blocks**
```F#
Program.mkProgram init update view
|> Program.withHost this
|> Program.runWithAvaloniaSyncDispatch ()
```

## 2) Breaking Changes to Subscriptions
* `Cmd.ofSub` has been removed
* `Program.withSubscription` now takes in the state/model, and returns a list of _named_ subscriptions.
* Subscription functions must now return `IDisposable` so they can be automatically disposed

The above changes will cause breakage in any FuncUI apps that make use of `Program.withSubscription`.

**Avalonia.FuncUI upgrade path**
Any FuncUI app that was previously using `Program.withSubscription` will have to resolve a build error since the function inputs/outputs have changed. 

**Old Elmish v3 Sample with Multiple Subscriptions**
```F#
        let timer (_state: Clock.State) =
            let sub (dispatch: Clock.Msg -> unit) =
                let invoke() =
                    DateTime.Now |> Clock.Msg.Tick |> dispatch
                    true
                    
                DispatcherTimer.Run(Func<bool>(invoke), TimeSpan.FromMilliseconds 1000.0) |> ignore
                
            Cmd.ofSub sub
        
        // Save state when host window is Closed
        let onClosedSub (state: Clock.State) = 
            Cmd.ofSub (fun _ -> 
                this.Closed.Subscribe(fun e -> 
                    printfn "Saving state: %A" state
                )
                |> ignore
            )

        Program.mkSimple (fun () -> Clock.init) Clock.update Clock.view
        |> Program.withHost this
        |> Program.withSubscription timer
        |> Program.withSubscription onClosedSub
        |> Program.withConsoleTrace
        |> Program.run
```

**New Elmish v4 Sample with Multiple Subscriptions**
```F#
        let subscriptions (_state: Clock.State) : Sub<Clock.Msg> =
            let timerSub (dispatch: Clock.Msg -> unit) =
                let invoke() =
                    DateTime.Now |> Clock.Msg.Tick |> dispatch
                    true
                    
                DispatcherTimer.Run(Func<bool>(invoke), TimeSpan.FromMilliseconds 1000.0) // <- IDisposable

            let onClosedSub (dispatch: Clock.Msg -> unit) =
                this.Closed.Subscribe(fun e -> printfn "The window has been closed.")

            [ 
                [ nameof timerSub ], timerSub
                [ nameof onClosedSub ], onClosedSub
            ]
        
        Program.mkSimple Clock.init Clock.update Clock.view
        |> Program.withHost this
        |> Program.withSubscription subscriptions
        |> Program.withConsoleTrace
        |> Program.run
```

(I'm not sure why each subscription identifier is a `string list` rather than a `string`.)

## Non-breaking Addition of `Program.runWith` Functions
Elmish v4 introduces the new `Program.runWith` style functions that allows the user to specify an arg for the `init` function.
`Program.run` still exists, and it assumes that `init` will take a `unit` arg. The new `runWith` functions allow the user to pass in the `init` arg. 

Just to be clear, `|> Program.run` is the equivalent of `|> Program.runWith ()`.

The main takeaway here is that there seems to be an assumption by the Elmish maintainers that `init` is expected to be a function that takes an arg (rather than making `init` a simple binding, as it is in many samples). 

This pattern is now used in the new `Program.runWith___` functions, such as `Program.runWithSyncDispatch`, which also expects the `init` arg to be passed.

I have updated the samples accordingly.

Benefits to this change:
* It adds consistency to always add an arg (even if it is usually `unit`) to the `init` function
* It removes the need to wrap `init`, and looks prettier. Example:
  * `Program.mkSimple (fun () -> init) update view` becomes:
  * `Program.mkSimple init update view`
* It better aligns with all the new `Program.runWith__` functions in Elmish 4.


sleepyfran (approved):
Awesome work! ❤️

JordanMarr:
Thanks Fran! 

Does anyone have any objections to me merging this into master now? I just wanted to check since it does introduce two breaking changes. 

Depending on the versioning strategy, I suppose this may warrant a major version bump. I think I will also need to add some sort of upgrade guidance.

sleepyfran:
I think we could go ahead and merge it. I'll take note about adding this stuff to a upgrade guide, we're going to need one anyway because there's been quite a lot of breaking changes in interfaces coming from the latest version of Avalonia so I think it should be safe 😃 

But of course @JaggerJo let us know what you think.

PR #249, opened 2023-01-17
More useElmish updates

JordanMarr:
# Overview of Changes

## Added Cmd.ofAsyncMsg to useElmish hook

Added a basic implementation of `Cmd.ofAsyncMsg`.
See [this](https://github.com/fsprojects/Avalonia.FuncUI/discussions/250) discussion.

Usage:

```F#
| OpenFiles ->
    let dialog = Dialogs.getMusicFilesDialog None
    let showDialog = 
        async {
            let! files = dialog.ShowAsync(window) |> Async.AwaitTask
            return AfterSelectFiles files
        }
    model, Cmd.ofAsyncMsg showDialog
```

## Breaking change to `useElmish` hook
* The `useElmish` hooks now expects `init` to take an `'arg`
* `useElmish` now has one overload that assumes the arg is `unit`, and another that allows the user to specify an `'arg`.

This is another breaking change: it requires that `init = ...` bindings will need to be refactored to `init() = ...`.

The benefit is that it better conforms to Elmish conventions for `init`, including the new [Fable 4](https://github.com/fable-compiler/Fable.React.UseElmish/blob/main/src/Fable.React.UseElmish/Fable.React.UseElmish.fs) `useElmish` hook which has similar overloads.

**Examples:**

Overload 1 (used when `init` takes a `unit` argument):
```F#
let init() = { Name = "" }

    let model, dispatch = ctx.useElmish(init, update)
```

Overload 2 (used when `init` takes a specific `'arg`):
```F#
let init(name: string) = { Name = name }

    let model, dispatch = ctx.useElmish(init, update, "Jordan")
```

## Updated `init` in the remaining samples
* Changed some samples that I missed before to use `init() = `

## Updated `init` in the templates
* Changed the templates to use `init()`

I also installed the templates locally to ensure that all four templates build properly.


JordanMarr:
Closing this and opening a new PR in its place.

PR #251, opened 2023-01-22
Complete rework of useElmish

JordanMarr:
# Overview of Changes

## Reworked useElmish hook
`useElmish` now actually uses Elmish 4 under the hood! 🎉 
This means that `useElmish` can now take advantage of all the great Elmish features like subscriptions and async commands!

It has two main variations of overloads: 
* Two that take `init`, `update`, `initArg` and an optional `mapProgram` (for adding subsriptions or other Elmish extensions)
* Two that take `writableModel`, `update` and an optional `mapProgram` (for adding subsriptions or other Elmish extensions)

Here is an example from the `ChordParserView` sample that uses subscriptions:

```F#
let private subscriptions (model: Model) : Sub<Msg> =
    let timerSub (dispatch: Msg -> unit) =
        let invoke() = dispatch Msg.SetTime; true
        DispatcherTimer.Run(invoke, TimeSpan.FromMilliseconds 1000.0)

    [ 
        // Dynamically start or stop (Dispose) subscription
        if model.Transpose = 0 then 
            [ nameof timerSub ], timerSub
    ]

    let model, dispatch = ctx.useElmish(init, update, (), Program.withSubscription subscriptions)
    //let model, dispatch = ctx.useElmish(init, update, ()) // if no subscriptions are needed
    
    Grid.create [
```

## More sample cleanup
* Changed some samples that I missed before to use `init() = ` instead of `init = `

## Updated `init` in the templates
* Same as above, but for the templates.
* I also installed the templates locally to ensure that all four templates build properly.


JordanMarr:
@sleepyfran This is good to go now... tested on the ChordParser example and ready to go. 👍

sleepyfran (approved):
Awesome work! Let's #shipit

JordanMarr:
@sleepyfran,
Looking at the build process, it looks like if I update the version it will automatically push the package to NuGet.
Is that correct? 

If so, then should I just create a new PR to update to the next preview version (0.6.0-preview6)?
Or can I just push directly to master?

Thanks!

sleepyfran:
> @sleepyfran,
> 
> Looking at the build process, it looks like if I update the version it will automatically push the package to NuGet.
> 
> Is that correct? 

Yep! The CI will automatically try to always push whatever version is defined, and then NuGet takes care of either rejecting it because there's already a version with the same number or accepting it if it's new 😄 

> If so, then should I just create a new PR to update to the next preview version (0.6.0-preview6)?
> 
> Or can I just push directly to master?

I think we don't have permissions to push to master directly, but feel free to create a PR and I will accept it :^)

PR #259, opened 2023-01-26
Elmish hook optional overloads

JordanMarr:
Added an overload to `useElmish` hook that makes `initArg` does not need to be specified when `initArg` is `unit`. 

So this:
```F#
let model, dispatch = ctx.useElmish(init, update, ())
```

Can also be written as this:
```F#
let model, dispatch = ctx.useElmish(init, update)
```


JordanMarr:
@sleepyfran, this one is ready, and I bumped the version to v0.6.0-preview7. 
It contains this change plus a render performance optimization (#237).
I will create a new release.

sleepyfran (approved):
Nice one, thanks!

PR #261, opened 2023-01-26
Use an object expression to implement IObservable rather than dependi…

Numpsy:
…ng on System.Reactive


Numpsy:
A test at spinning off part of #246 just to try removing a local system of System.Reactive.Linq without updating Avalonia

PR #267, opened 2023-02-05
Attached properties related to text input

Happypig375:


JaggerJo:
LGTM

Could you explain why they are in `StyledElement.fs` ? / Would it make sense to put them in a separate file? 

(Really don't know how they are related to `StyledElement`, so I don't have an opinion yet)

JaggerJo:
LGTM

Could you explain why they are in `StyledElement.fs` ? / Would it make sense to put them in a separate file? 

(Really don't know how they are related to `StyledElement`, so I don't have an opinion yet)

JaggerJo:
LGTM

Could you explain why they are in `StyledElement.fs` ? / Would it make sense to put them in a separate file? 

(Really don't know how they are related to `StyledElement`, so I don't have an opinion yet) #

JaggerJo (approved):

Happypig375:
@JaggerJo The attached properties are applicable to StyledElements by looking at Avalonia source code here.

https://github.com/AvaloniaUI/Avalonia/blob/38fb9f18c463abb39eedd49e3f97004d6f7dc6f5/src/Avalonia.Base/Input/TextInput/TextInputOptions.cs

They could technically be put in another file but I feel that having all StyledElement properties together makes finding the source code easier.

JaggerJo:
> @JaggerJo The attached properties are applicable to StyledElements by looking at Avalonia source code here.
> 
> https://github.com/AvaloniaUI/Avalonia/blob/38fb9f18c463abb39eedd49e3f97004d6f7dc6f5/src/Avalonia.Base/Input/TextInput/TextInputOptions.cs
> 
> They could technically be put in another file but I feel that having all StyledElement properties together makes finding the source code easier.

Ok, makes sense. Agree 👍

PR #272, opened 2023-02-16
Add a WebAssembly control catalog project

Numpsy:
Something that I was having a play with out of personal interest, that I thought i'd put here in case anyone is interested - I rearranged the control catalog so that all the views etc are in their own project and then added a web assembly project (self contained, rather than Blazor based) that hosts that along side the existing desktop app.

The control catalog in Avalonia itself has something similar.

It sppears to be working as long as I don't try to debug it, though the startup performance is sometimes quite slow:

image

The changes here aren't actually that large - most of the diff is from moving the existing files down into an additional subdirectory, so that the Desktop/Browser apps can go along side it.


JaggerJo (approved):
LGTM

Numpsy:
Ok.
Looks like the build failure is because there is no workload restore step in the PR build script (though there is one in the publish pipeline)

PR #277, opened 2023-03-01
Create a useStateLazy method which takes a callback

marklam:
This means that if your default state is expensive to create, you don't need to create one (just to be discarded) every time the component is rendered.

### DIFF
--- a/src/Avalonia.FuncUI/Components/Context/Context.fs
+++ b/src/Avalonia.FuncUI/Components/Context/Context.fs
@@ -53,6 +53,24 @@ type IComponentContext =
     abstract useState<'value> : initialValue: 'value * ?renderOnChange: bool -> IWritable<'value>
 
     /// <summary>
+    /// Attaches a value from a function to the component context.
+    /// <example>
+    /// <code>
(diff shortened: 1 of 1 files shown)


src/Avalonia.FuncUI/Components/Context/Context.fs
```diff
@@ -52,6 +52,24 @@ type IComponentContext =
     /// <param name="renderOnChange">re-render component on change (default: true).</param>
     abstract useState<'value> : initialValue: 'value * ?renderOnChange: bool -> IWritable<'value>
 
+    /// <summary>
+    /// Attaches a value from a function to the component context.
+    /// <example>
+    /// <code>
+    /// Component (fun ctx ->
+    ///     let count = ctx.useStateLazy (fun () -> 42)
```
JaggerJo, line 60:
```suggestion
    ///     (* expensive operation that should only happen once on initialisation *) 
    ///     let count = ctx.useStateLazy (fun () -> Math.Sqrt(42))
```

JaggerJo (approved):
LGTM

PR #278, opened 2023-03-04
Contexts / Environment state

JaggerJo:
Allows controls to pass state to descendants (similar to swift UI environment objects & react contexts)


JaggerJo:
Quite impressive how far you get with [~80 LOC](https://github.com/fsprojects/Avalonia.FuncUI/pull/278/files#diff-5c284004983cedb08f295b46ac76f1513feef63be5e6de673ed9dfc196f1d039) which is almost the entire feature.

JaggerJo:
@JordanMarr @sleepyfran  I'm not happy with the naming and the API yet, but it works. Feedback is as always appreciated.

### Describe value

https://github.com/fsprojects/Avalonia.FuncUI/blob/e2ddebad32c6edd13e6b5afd52ca5840a614168b/src/Examples/Component%20Examples/Examples.EnvApp/Program.fs#L15-L19

### Provide value

https://github.com/fsprojects/Avalonia.FuncUI/blob/e2ddebad32c6edd13e6b5afd52ca5840a614168b/src/Examples/Component%20Examples/Examples.EnvApp/Program.fs#L172-L197

### Consume value

https://github.com/fsprojects/Avalonia.FuncUI/blob/e2ddebad32c6edd13e6b5afd52ca5840a614168b/src/Examples/Component%20Examples/Examples.EnvApp/Program.fs#L24-L32

sleepyfran:
Looks good, although I'd maybe make the whole providing a bit more friendly by going full React. If you make the `EnvironmentState` contain a function that creates the Provider automatically, the providing part would end up like:

```fsharp
SharedState.brush.provider(
             value = brush, 
             content = (
                 ...
             )
)
```

Consumption-wise I think the API is okay. 

I'm not a huge fan of contexts in general because they couple components with an invisible provider that makes it very hard to refactor later (going right now through a very painful transition with them in React at work, so this hits close to home 😆), and I think we have this partially solved with the store pattern that we have in some of the samples, but I also think they won't hurt and it might feel more familiar for people coming from React.

JordanMarr:
I haven’t quite recovered from my GitHub session yesterday, but I will look at this in more detail later. 

@sleepyfran I recently refactored a React app to use Fable.Store from useContext. What is an example of the store pattern you are referring to for FuncUI?

sleepyfran:
@JordanMarr for example the Generic Algorithm sample declares this store:

https://github.com/fsprojects/Avalonia.FuncUI/blob/200bae3ee10a435b2a40c4e2ef40a4b857e21e3f/src/Examples/Component%20Examples/Examples.GeneticAlgorithm/Views.fs#L15-L28

Which are then used as a normal state:

https://github.com/fsprojects/Avalonia.FuncUI/blob/200bae3ee10a435b2a40c4e2ef40a4b857e21e3f/src/Examples/Component%20Examples/Examples.GeneticAlgorithm/Views.fs#L168-L170

I'm not familiar with Fable.Store, so not sure if they do something different, but basicaly this just creates a global state that is then shared between all components that reference it. So in order to have something similar to a context, declaring a store like this scoped to a set of components would end up being the same.

JaggerJo:
> Looks good, although I'd maybe make the whole providing a bit more friendly by going full React. If you make the `EnvironmentState` contain a function that creates the Provider automatically, the providing part would end up like:
> 
> ```fsharp
> SharedState.brush.provider(
>              value = brush, 
>              content = (
>                  ...
>              )
> )
> ```
> 
> Consumption-wise I think the API is okay.

Good point 👍 that might be handy.

> I'm not a huge fan of contexts in general because they couple components with an invisible provider that makes it very hard to refactor later (going right now through a very painful transition with them in React at work, so this hits close to home 😆), and I think we have this partially solved with the store pattern that we have in some of the samples, but I also think they won't hurt and it might feel more familiar for people coming from React.

I personally think both (global shared state & shared state via contexts) have their place. 
If state only starts to exist at some point contexts make a lot of sense.

sleepyfran:
> If state only starts to exist at some point contexts make a lot of sense.  
> 
> 
> 
>

That's actually a really good point. I guess I've never had this exact use case, but it definitely makes sense to offer this out of the box.

Well, as I said, other than the provider change the rest looks all good to me. Awesome job!

PR #282, opened 2023-03-15
Update for latest Avalonia 11 changes

Numpsy:
The thread at https://github.com/AvaloniaUI/Avalonia/discussions/10653#discussioncomment-5316935 says there is a new preview release coming soon, and of course there are more breaking changes in the current nightly builds, so changes ho..

The changes are mainly flyout related stuff due to the change at https://github.com/AvaloniaUI/Avalonia/pull/10492

There are also a few WASM changes due to the startup functions having been changed.

Note; The WASM control catalog builds with this change for me, but I get a stack overflow in the browser when I run it - I don't know ATM if that's just something missing over here, or something is wrong in the nightly builds


src/Examples/Component Examples/Examples.WebAssemblyPlayground/Startup.fs
```diff
@@ -33,5 +24,14 @@ module Program =
 
         let builder = WebAssemblyHostBuilder.CreateDefault(args)
         builder.RootComponents.Add<MainView>("#app")
-        builder.Build().RunAsync() |> ignore
+        let host = builder.Build()
+
+        task {
+            do! AppBuilder.Configure<App>()
+                      .UseReactiveUI()
+                      .StartBlazorAppAsync()
```
Numpsy, line 32: The WASM startup functions in Avalonia have been made async, and i'm not sure what the best way of handling those is (it currently just ignores the created task like the previous code did)

JaggerJo:
@Numpsy nice! Do you mind if we postpone merging this until it's in a preview instead of nightly?

Numpsy:
That's one of the reasons why it's a draft :-)

Actually, I see another pending change at https://github.com/AvaloniaUI/Avalonia/pull/10664 that renames some more things, so best to let that settle anyway I think

JaggerJo:
Yeah, guess it makes sense to wait this out :D btw. I've sent you an email.

Numpsy:
It looks like the stack overflow I was getting is an Avalonia bug introduced since the last preview release - https://github.com/AvaloniaUI/Avalonia/issues/10675

Numpsy:
A note / question on the change to ```ItemsControl.dataItems``` - 

The Avalonia changes at 
https://github.com/AvaloniaUI/Avalonia/pull/10590
https://github.com/AvaloniaUI/Avalonia/pull/10604

Have
1) Added an ```ItemsSource``` property to ItemsControl
2) Changed ```ItemsControl.Items``` from an ```IEnumerable``` to ```IList```

Based on the doc comments in those changes:
image
I changed ```dataItems``` to bind to ```ItemsSource``` rather than changing the type of the property to IList.

That has mostly got the control catalog working, except for the tree view sample, which fails at runtime with 
```
An unhandled exception of type 'System.ArgumentException' occurred in Avalonia.Base.dll
Invalid value for Property 'Items': '[{ Name = "Fruit"
  Children = [{ Name = "Tomato"
                Children = [] }; { Name = "Apple"
                                   Children = [] }] }; { Name = "Vegetables"
  Children = [{ Name = "Carrot"
                Children = [] }; { Name = "Salad"
                                   Children = [] }] }]' (Microsoft.FSharp.Collections.FSharpList`1[[Avalonia.FuncUI.ControlCatalog.Views.TreeViewDemo+Taxonomy, Avalonia.FuncUI.ControlCatalog, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]])
```

As it appears that something is trying to convert the provided IEnumerable (which is an FSharpList) to an IList (it works if the Children collections in the sample are arrays instead of lists, and the root collection being a list seems to  just work).

I'm not sure if this is a bug somewhere else, or just something in the way that something is populating the 'Items' collection in the child treeview nodes, but it makes me wonder if the ```itemsSelector``` part of DataTemplateView might need to be an ```IList``` instead of a ```seq``` ?

JaggerJo:
Good questions. Need to take a deeper look myself.

Numpsy:
Updated to the Avalonia 11.0.0-preview6 release build.

The web assembly bits seem to have been fixed now.

Numpsy:
So, I'd been looking at that items control code I linked and wondering if it should bind to ItemsSource rather than Items, and now I see a PR at https://github.com/AvaloniaUI/Avalonia/pull/10827 that does just that (though it also makes the Items property read only, so I'm sure that will break something else).

Anyway, for the Avalonia preview 6 release all my stuff seems to work with this PR as long as I only return ILists from my tree template code, and it doesn't look like we'd want to change the public FuncUI interface to an IList now if IEnumerable will work again later... so - what's the best way to proceed with the other changes?

JaggerJo (approved):
LGTM, feel free to bump the version and merge.

PR #283, opened 2023-03-15
Add a binding for ToggleButton.IsCheckedChangedEvent

Numpsy:
refs this deprecation in Avalonia 11:

image

where ```IsCheckedChangedEvent``` is a new one.

This is perhaps a tad confusing in the bindings, where there is already an ```onIsCheckedChanged``` bound to ```IsCheckedProperty``` which is itself marked as obsolete, so not sure what the desired approach is there.

### DIFF
--- a/src/Avalonia.FuncUI/DSL/Buttons/ToggleButton.fs
+++ b/src/Avalonia.FuncUI/DSL/Buttons/ToggleButton.fs
@@ -23,10 +23,9 @@ module ToggleButton =
             
         static member isChecked<'t when 't :> ToggleButton>(value: bool option) : IAttr<'t> =
             value |> Option.toNullable |> ToggleButton.isChecked
-            
-        [<Obsolete "use 'onChecked' or 'onUnchecked' instead">]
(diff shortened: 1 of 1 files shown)


src/Avalonia.FuncUI/DSL/Buttons/ToggleButton.fs
```diff
@@ -27,6 +27,9 @@ module ToggleButton =
         [<Obsolete "use 'onChecked' or 'onUnchecked' instead">]
         static member onIsCheckedChanged<'t when 't :> ToggleButton>(func: Nullable<bool> -> unit, ?subPatchOptions) : IAttr<'t> =
```
JaggerJo, line 28: I think we should get rid of this method avoid confusion.

JaggerJo (approved):
LGTM, but remove the old method before merging.

src/Avalonia.FuncUI/DSL/Buttons/ToggleButton.fs
Numpsy: done

PR #289, opened 2023-04-02
data grid bindings

JaggerJo:


JordanMarr (approved):
Very nice! A much needed addition. 👍

JaggerJo:
Yeah, it's about time 😄

The bindings aren't complete yet - somehow the runtime even crashes when double clicking a text cell 😅

Adding the bindings to the current project also means adding a reference to `Avalonia.Controls.DataGrid`.

Numpsy:
> Adding the bindings to the current project also means adding a reference to `Avalonia.Controls.DataGrid`

We do at least seem to be getting to a point where the linker trimming machinery can remove the extra libraries at build time if they aren't actually being called :-)

JordanMarr:
> Yeah, it's about time 😄
> 
> The bindings aren't complete yet - Avalonia even [crashes when double clicking a text cell](https://github.com/AvaloniaUI/Avalonia/issues/10843) 😅

Interesting... it doesn't crash for me when I double click the text cell. 🤷‍♂️

> Adding the bindings to the current project also means adding a reference to `Avalonia.Controls.DataGrid`.

That seems ok to me since the DataGrid is such a commonly used control in most projects.

PR #130, opened 2020-07-29
Add Drag & Drop and port the Avalonia Drag+Drop demo to FuncUI

renngar:
This adds Drag & Drop support to FuncUI and ports the C# Avalonia Drag+Drop demo to F#.

As mentioned on gitter, when you click on the drag source, onPointerPressed it triggered twice.  I am uncertain if this is a bug in my code, or a bug in Avalonia or FuncUI.

The final commit below has an ugly hack to workaround this.

PR #295, opened 2023-04-07
Drag and Drop functionality

Numpsy:
This contains the changes from #130 plus

1) It's rebased on top of the current master branch to pick up changes like the DSL code being moved into the main lib and the Elmish v4 updates
2) There's an extra commit to update it to deal with a couple of Avalonia 11 changes


JaggerJo (approved):
LGTM

JaggerJo:
LGTM, should I merge?

Numpsy:
The demo in the control catalog does seem to be working ok with the latest code.

(I was previously intending to try adding drop-drop file opening to a tool I was working on that would use this, but I've been snowed under with other things and haven't got to it yet :-( )

Numpsy:
I tried plugging it into a test app and it works sufficiently to open a file via drag and drop, so the base functionality looks good :-)

dragdrop

PR #301, opened 2023-04-19
New properties in Avalonia supporting touch scrolling

marklam:
A couple of new properties on ItemsControl related to scrolling (for touch UI, but the size calculations they affect are performed for all UI).


JaggerJo:
LGTM

PR #303, opened 2023-04-25
[wip] Start updating to Avalonia 11 preview 7

Numpsy:
A start on getting things working with preview 7.

main changes:

-The classes property is now read only
-Various controls have been changed to use ItemsSource instead of Items
-The Items property on ItemsControl has been made a regular instance property, which is read only

I tried changing ItemsControl.viewitems but haven't changed onItemsChanged and there is also a change needed to clipboard access in the diagnostics lib

### DIFF
--- a/src/Avalonia.FuncUI.Diagnostics/Views/Views.StateHookViews.fs
+++ b/src/Avalonia.FuncUI.Diagnostics/Views/Views.StateHookViews.fs
@@ -24,12 +24,12 @@ type internal StateHookViews =
                         ContextMenu.viewItems [
                             MenuItem.create [
                                 MenuItem.header "Copy"
-                                MenuItem.onClick (fun _ ->
+                                MenuItem.onClick (fun args ->
                                     Async.StartImmediate (
                                         async {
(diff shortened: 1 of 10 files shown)


Numpsy:
On a related note there are also piles of warnings from inside the library about ViewBuilder.Create being obsolete - do those all need updating?

JaggerJo:
> On a related note there are also piles of warnings from inside the library about ViewBuilder.Create being obsolete - do those all need updating?

You can remove the obsolete attribute. The `ViewBuilder.Create` existed in two places, so this one showed up as having no usages. As we have a better API exposed `View.create...` I thought we don't need this one anymore - apparently that was a mistake.

Numpsy:
So, didn't get chance to work on this for a bit, and now there is a preview 8 as well :-(

The remaining failure on this set of changes was that the Items property on ItemsControl is no longer an avalonia property, it's now just a plain class property (of type ItemCollection), so the onItemsChanged binding doesn't work. 
ItemCollection does seem to implement INotifyCollectionChanged and similar events, so maybe it needs to be reworked to use those?

PR #304, opened 2023-05-12
Preview 8

JaggerJo:


JaggerJo:
@Numpsy everything builds now 🎉

Numpsy:
I wasn't clear about what the ItemsControl itmes changed event should be bound to, as I thought I'd read that it can be used with an items collection without setting ItemsSource to a value, but i'm not entirely clear on the relationship between the two :-(

On a positive note, the menu flyout example in the control catalog app was showing an empty dropdown with the preview 6 build, and that looks to have been fixed in the update

JaggerJo:
> I wasn't clear about what the ItemsControl itmes changed event should be bound to, as I thought I'd read that it can be used with an items collection without setting ItemsSource to a value, but i'm not entirely clear on the relationship between the two :-(

This fires every time the button is clicked, but as you say it does not work if `viewItems` is used instead of `dataItems`.

```fsharp
Component (fun ctx ->
    let toggle = ctx.useState false
    let items = ctx.useState<IEnumerable> [ 0 .. 10 ]
    DockPanel.create [
        DockPanel.children [
            Button.create [
                Button.dock Dock.Bottom
                Button.onClick (fun _ ->
                    if toggle.Current then
                        items.Set [ 0 .. 10 ]
                    else
                        items.Set [ 'A' .. 'C' ]
                    toggle.Set (not toggle.Current)
                )
                Button.content "change items"
                Button.horizontalAlignment HorizontalAlignment.Stretch
            ]
            ItemsControl.create [
                ItemsControl.onItemsChanged (fun _ -> printfn "Items changed!")
                ItemsControl.dataItems items.Current
            ]
        ]
    ]
)
```


> On a positive note, the menu flyout example in the control catalog app was showing an empty dropdown with the preview 6 build, and that looks to have been fixed in the update

Nice!

PR #307, opened 2023-05-29
Updated Templates

AngelMunoz:
This PR adds the updated templates from both FuncUI and Avalonia and also removed the JaggerJo prefix from readmes and files in case these are pushed to the Avalonia.FuncUI namespace in NuGet

The basic templates both work fine but the "full" ones don't work, since we use the tabs control and that uses `viewItems` property which seems to have been a breaking change from up stream but I think we're aware of that due conversations in #303 and #304

```fs
Unhandled exception. System.MissingFieldException: Field not found: 'Avalonia.Controls.ItemsControl.ItemsProperty'.
   at Avalonia.FuncUI.DSL.ItemsControl.ItemsControl.viewItems........
```

since those changes haven't been released yet I can't test them with these templates but once that is fixed and released these should be able to go as well


Numpsy:
I think those ItemsControl changes in Avalonia should have been handled now (the related issues in the control catalog are fixed anyway)

AngelMunoz:
Yes they are in the master branch, but they haven't been release yet, now that Avalonia has announced the RC we can wait to update these templates to FuncUI RC version as well once it is released!

AngelMunoz:
I just updated the templates' versions to the ones specified in `src\Directory.Build.props` for both Avalonia and FuncUI
I'll keep an eye for more version updates

JaggerJo:
Nice to see the interest in templates. Thanks for all the work. 

I think if we publish templates we should make sure they are always up to date and working. 

Is it possible to
- use the latest version of FuncUI specified in `Directory.props`
- "build" the templates so that we are sure they will at least compile

JaggerJo:
I'm going to merge this in the current state. 

I'll make sure the templates are tested and minimal.

AngelMunoz:
Hey there sorry, I wasn't around yesterday on GH


> "build" the templates so that we are sure they will at least compile

I think the Main Avalonia templates have a powershell script setup to build out templates just like you mention I will explore that

> use the latest version of FuncUI specified in Directory.props


I'm not aware of that as they are installed in the user's machine and I'm not sure if that prop is available at that time, do you happen to know @Happypig375 ?

JaggerJo:
> Hey there sorry, I wasn't around yesterday on GH
> 
> > "build" the templates so that we are sure they will at least compile

No worries, I also want to take a closer look myself. 

> I think the Main Avalonia templates have a powershell script setup to build out templates just like you mention I will explore that
> 
> > use the latest version of FuncUI specified in Directory.props
> 
> I'm not aware of that as they are installed in the user's machine and I'm not sure if that prop is available at that time, do you happen to know @Happypig375 ?

What I mean here is that if we release a new version of FuncUI I want to automatically release a new version of the templates targeting that new version.

PR #311, opened 2023-06-19
Restored the iOS sample in the solution

Thecentury:
Previously, the solution contained the `Examples.Mobile` folder but there were no projects there.


Thecentury:
I see that the build fails with the error
```
Unhandled exception: Microsoft.Build.Exceptions.InvalidProjectFileException: The imported project "/home/runner/.dotnet/sdk/7.0.304/Sdks/Microsoft.NET.Sdk/16.0.1478/targets/Xamarin.Shared.Sdk.MultiTarget.targets" was not found. Confirm that the expression in the Import declaration ";../16.0.1478/targets/Xamarin.Shared.Sdk.MultiTarget.targets" is correct, and that the file exists on disk.  /home/runner/.dotnet/sdk/7.0.304/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets
```

Is that a reason why the project is excluded from the solution?

Numpsy:
does it need to install an ios workload onto the build agent or soemthing like that?

JaggerJo:
After a lot of tries I got it working in CI.

PR #326, opened 2023-06-30
Update to Avalonia 11.0.0-rc2.2

Numpsy:
Just starting to update ready for the next version whilst I've got a spare minute.

Some functions were removed from ItemsControl is https://github.com/AvaloniaUI/Avalonia/pull/11752, and I noticed that the same properties also exist on StackPanel and didn't have bindings, so I added them at the same time


Numpsy:
This seems to work for me locally when I build with ```dotnet build``` or in VSCode, but in VS 2022 I get https://github.com/AvaloniaUI/Avalonia/issues/11913

Numpsy:
Updated to 11.0.0-rc2.2

JaggerJo:
LGTM

JaggerJo (approved):

PR #335, opened 2023-07-24
Bump versions on the templates

AngelMunoz:
I haven't had a chance to come up with a more sophisticated way to check out that the templates build fine other than do it on CI by script so I'll add that in the meantime

I Omitted the cross-platform one because I currently can't test it locally in my machine and in case of failure I wouldn't be able to debug

pwsh is available in github actions so it should work just fine

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,9 @@ jobs:
       
     - name: Build & Test
       run: dotnet test src/Avalonia.FuncUI.sln --verbosity normal
+
+    - name: Build Templates
+      run: pushd ./src/Avalonia.FuncUI.Templates && ./test-templates.ps1
       
     - name: Build for release
       run: dotnet build src/Avalonia.FuncUI/Avalonia.FuncUI.fsproj -c Release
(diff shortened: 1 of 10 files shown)


JaggerJo:
Will take a look after my vacation

Numpsy:
Is anything happening with these template changes?

JaggerJo:
Personally I don't like or use templates.

If we publish them someone need to keep them up to date, make sure they still build, ..

I never merged them because that's not something I'd like to do. If someone else volunteers to keep them up to date or if we can completely automate it I'm all for it.

AngelMunoz:
@Numpsy Welp, I had them ready to merge a while ago but it never happened, I was willing to keep them up to date if they were accepted in the first place but that is not the case, I'm no longer interested in putting time and effort in maintaining them.

I think I left them ready for v11 and I think there haven't been breaking changes in FuncUI so they should still work.

Feel free to close this if there's no interest in them.

Numpsy:
> Personally I don't like or use templates.
> 
> If we publish them someone need to keep them up to date, make sure they still build, ..
> 
> I never merged them because that's not something I'd like to do. If someone else volunteers to keep them up to date or if we can completely automate it I'm all for it.

I'm not using them for simple projects myself.
I think the templates are more use for things like web assembly where there are more moving parts (javascript code etc as well as .NET), but then Avalonia itself has F# templates for that, and just using those and then inserting FuncUI afterwards might be easier than maintaining a local template anyway.

PR #346, opened 2023-09-06
Allow IComponentContext.useElmish with supplied mapProgram to change msg type

marklam:
From a question I raised in https://github.com/fsprojects/Avalonia.FuncUI/discussions/339

It's possible to change the msg type if the useElmish method is split into overloads which take / don't take mapProgram - rather than an optional parameter which defaults to `id`

I've added the somewhat pointless `withTimings` to the Chord Parser app just to demonstrate it working.

If you'd rather not merge it, I have a workaround which sits outside Avalonia.FuncUI, but I thought I'd offer it up anyway.

### DIFF
--- a/src/Avalonia.FuncUI.Elmish/ElmishHook.fs
+++ b/src/Avalonia.FuncUI.Elmish/ElmishHook.fs
@@ -2,9 +2,10 @@
 
 open Avalonia.FuncUI
 open Elmish
+open System
 
 /// Starts an Elmish loop and provides a Dispatch method that calls the given setModel fn.
-type ElmishState<'model, 'msg, 'arg>(mkProgram : unit -> Program<'arg, 'model, 'msg, unit>, arg: 'arg, setModel) =
+type ElmishState<'model, 'msg, 'msg2, 'arg>(mkProgram : unit -> Program<'arg, 'model, 'msg, unit>, mapProgram : Program<'arg, 'model, 'msg, unit> -> Program<'arg, 'model, 'msg2, unit>, arg: 'arg, setModel) =
     let program = mkProgram()
 
     let mutable _model = Program.init program arg |> fst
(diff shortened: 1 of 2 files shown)


JaggerJo:
@JordanMarr could you take a look?

JordanMarr:
I'm not sure that implementation will handle the scenario where you map other things, like the view or the model.
I will kick it around a bit and see if I can come up with a solution.

JordanMarr:
I have to admit that I still don't fully understand your use case, but it seems very specific to your project. (It doesn't help that I could find no examples of `Program.map` in the wild to help inform my intuition on how it is commonly used.)
My gut feeling is that it's not a feature that will be needed by others, and so I don't think the added complexity is worth it. Especially considering that fact that you can easily create your own extensions methods within your project (as you have probably done).

edgarfgp:
Looks similar to what we have in Fabulous that allows to map a `CmdMsg` into a regular `Msg` 🤔https://github.com/fabulous-dev/Fabulous.Avalonia/blob/b32481fe7d6d3cc62ca9501656a919d6c8228701/src/Fabulous.Avalonia/Program.fs#L109

marklam:
That's fair. I found a way round it in about the same amount of code.  

I'll close this PR and maybe comment my workaround in the linked discussion when I get a moment.

JordanMarr:
After experimenting with it, I have something that I could live with, but it works slightly differently than what you had.
The difference would be that it returns a `dispatch: (TimedMsg -> unit)`, so you would still need to map the `Msg`:

```F#
    let model, dispatch = ctx.useElmish(init, update,  Program.withSubscription subscriptions >> withTimings)
    let dispatch = TimedMsg.map >> dispatch
```

```F#
type TimedMsg = { Time: DateTime; Message: Msg } with
    static member map msg = { Time = DateTime.UtcNow; Message = msg }
```

The implementation is here: 
https://github.com/fsprojects/Avalonia.FuncUI/compare/master...JordanMarr:Avalonia.FuncUI:elmish-hook-map-program

Would that be flexible enough to work for your situation?

marklam:
Yes, I think that would work ok.

PR #362, opened 2023-10-30
Fix example titles

beyon:
A couple of the example projects have the same window title as Examples.DrawingApp which can lead to confusion when looking at the examples. Maybe the titles can be even better but at least this fixes the ones that are incorrect.


JaggerJo (approved):
LGTM

PR #363, opened 2023-10-30
Adding more Color overloads for brush properties

beyon:
Before: Some properties had overloads taking a Color value in addition to IBrush and a color string.
This commit adds overloads for brush properties missing a Color overload making it more consistent and ergonomic since at least in my mind setting the brush to a solid color is probably the most common scenario.

I was tempted to also change the parameter names in some cases where they are the generic term "value" to "color"/"brush" but then refrained since it could possibly break someone's code if they use named parameters and also not sure what you think of such a change.


JaggerJo (approved):
LGTM, thanks for the contribution.

JaggerJo:
It would surprise me if the parameter names were used.

PR #364, opened 2023-10-31
Gridsplitter properties

beyon:
This commit adds missing properties to GridSplitter, before no properties were implemented. It also adds GridSplitter Demo to the Controls Catalog.

GridSplitter.showsPreview seems to not be fully working in Avalonia at the moment (https://github.com/AvaloniaUI/Avalonia/issues/10700)

I am not sure how to test/demo GridSplitter.contentPreview since it's hard to find examples and documentation of usage but I think the types should be correct.

The GridSplitter demo allows you to play around with and test all the properties except .contentPreview


JaggerJo (approved):
LGTM, Thanks for the contribution 👍

Issue #365, opened 2023-10-31
Consider extending the DSL for more convenient ColumnDefinitions and RowDefinitions


beyon:
## Motivation/Background:
Setting minWidth and/or minHeight on the "cells" directely of a Grid doesn't work in Avalonia, at least not currently (see: https://github.com/AvaloniaUI/Avalonia/issues/7637).

You can't specify minHeight and minHeight when specifying rowDefinition/columnDefinition in string form, instead you have to use ColumnDefinitions and RowDefinitions objects instead.

Using the ColumnDefinitions, RowDefinitions, ColumnDefinition, RowDefinition objects in FuncUI results in verbose code that doesn't fit so nicely with the FuncUI DSL.

It's also nice to not have an alternative to specifying the definitions in a string with no validation.

I ended up writing my own DSL code when working around the minimum size issue and made a demo project: https://github.com/beyon/FuncUIGridCellMinWidthAndHeight

More information is available in the demo project readme file.

Current way of working with ColumnDefinitions and RowDefinitions:
```fsharp
    //Column definitions
    let col1 = ColumnDefinition()
    let col2 = ColumnDefinition()
    let col3 = ColumnDefinition()
    col1.Width <- GridLength(1.0, GridUnitType.Star)
    col1.MinWidth <- 64.0
    col2.Width <- GridLength(4.0, GridUnitType.Pixel)
    col2.MinWidth <- 4.0
    col3.Width <- GridLength(3.0, GridUnitType.Star)
    col3.MinWidth <- 64.0
    let colDefs = ColumnDefinitions()
    colDefs.Add(col1)
    colDefs.Add(col2)
    colDefs.Add(col3)
    //Row definitions
    let row1 = RowDefinition()
    let row2 = RowDefinition()
    let row3 = RowDefinition()
    row1.Height <- GridLength(1.0, GridUnitType.Star)
    row1.MinHeight <- 64.0
    row2.Height <- GridLength(4.0, GridUnitType.Pixel)
    row2.MinHeight <- 4.0
    row3.Height <- GridLength(1.5, GridUnitType.Star)
    row3.MinHeight <- 64.0
    let rowDefs = RowDefinitions()
    rowDefs.Add(row1)
    rowDefs.Add(row2)
    rowDefs.Add(row3)
    Grid.create [
        Grid.columnDefinitions colDefs
        Grid.rowDefinitions rowDefs
        Grid.children [
                ...
```
## Possible solution
[...]

PR #371, opened 2023-11-15
Adding ColumnDefinition and RowDefinition to DSL

beyon:
This implements what was discussed in #365 

The hardest part is naming things, I ended up with duplicate DU:s (same DU cases) for the sizing called ColumnWidth and RowHeight as I had trouble coming up with a good name fitting both usages. GridSize was probably my best alternative but it reads like it applies to the Grid as a whole, I would be ok with changing to that though.

### DIFF
--- a/src/Avalonia.FuncUI/DSL/Panels/Grid.fs
+++ b/src/Avalonia.FuncUI/DSL/Panels/Grid.fs
@@ -1,6 +1,68 @@
 namespace Avalonia.FuncUI.DSL
 
 [<AutoOpen>]
+module ColumnDefinition =  
+    open Avalonia.Controls
+    
+    [<RequireQualifiedAccess>]
+    type ColumnSize =
+    /// The column is auto-sized to fit it's contents
+    | Auto
+    /// The column is sized in device independent pixels
+    | Pixel of float
+    /// The column is sized as a weighted proportion of available space
+    | Star of float
+
+    type ColumnDefinition with
+
+        /// <summary>
+        /// Define the properties of a Grid column
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Avalonia.FuncUI/DSL/Panels/Grid.fs
```diff
@@ -1,5 +1,67 @@
 namespace Avalonia.FuncUI.DSL
 
+[<AutoOpen>]
+module ColumnDefinition =  
+    open Avalonia.Controls
+    
+    [<RequireQualifiedAccess>]
+    type ColumnWidth =
```
JaggerJo, line 8: Can we make this DU a `[<Struct>]` ?

src/Avalonia.FuncUI/DSL/Panels/Grid.fs
```diff
@@ -1,5 +1,67 @@
...
+            minWidth |> Option.iter (fun minW -> columnDefinition.MinWidth <- minW)
+            maxWidth |> Option.iter (fun maxW -> columnDefinition.MaxWidth <- maxW)
+            columnDefinition
+
+[<AutoOpen>]
+module RowDefinition =  
+    open Avalonia.Controls
+
+    [<RequireQualifiedAccess>]
+    type RowHeight =
```
JaggerJo, line 39: Can we make this DU a `[<Struct>]` ?

JaggerJo (approved):

src/Avalonia.FuncUI/DSL/Panels/Grid.fs
beyon: Yes can't think of any drawbacks and probably better as a value type.

PR #374, opened 2023-11-21
Replace a use of Seq.iteri over a list with List.iteri

Numpsy:
I've been having a go with with the [F# analyzers](http://ionide.io/FSharp.Analyzers.SDK/) tools, and got this suggestion that using List.iteri rather than Seq.iteri when the collection is known to be a list should be more efficient

```
Avalonia.FuncUI\VirtualDom\VirtualDom.Patcher.fs(92,25): Warning GRA-VIRTUALCALL-001 - Consider replacing the call of Seq.iteri with a function from the List module to avoid the costs of virtual calls.
```


JaggerJo:
LGTM

Issue #381, opened 2024-01-08
Add PathIcon Control


Miticcio:
I noticed that the [PathIcon](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/PathIcon.cs) control is missing in the DSL.
Is this intentional or can it be added?

I tried including it in a local build and it works fine, it is similar to Shapes.Path

`PathIcon.fs:`
```F#
namespace Avalonia.FuncUI.DSL

[<AutoOpen>]
module PathIcon =
    open Avalonia.Media
    open Avalonia.FuncUI.Builder
    open Avalonia.FuncUI.Types
    open Avalonia.Controls
    
    let create (attrs: IAttr<PathIcon> list): IView<PathIcon> =
        ViewBuilder.Create<PathIcon>(attrs)
     
    type PathIcon with

        static member data<'t when 't :> PathIcon>(geometry: Geometry) : IAttr<'t> =
            AttrBuilder<'t>.CreateProperty<Geometry>(PathIcon.DataProperty, geometry, ValueNone)
            
        static member data<'t when 't :> PathIcon>(data: string) : IAttr<'t> =
            data |> Geometry.Parse |> PathIcon.data
```

JaggerJo:
Happy to merge this if you create a PR

PR #382, opened 2024-01-08
Added PathIcon Control

Miticcio:
https://github.com/fsprojects/Avalonia.FuncUI/issues/381


Numpsy:
@JaggerJo Does the change in #380 need to be made to the PR workflow? The failure here sounds the same as that (i.e. not actually related to the change here)

PR #383, opened 2024-01-08
Use 'dotnet workload restore' to restore workloads in the PR CI builds

Numpsy:
refs #380 and the build failure in #381

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -22,11 +22,8 @@ jobs:
           6.x.x
           7.x.x
 
-    - name: install workload ios
-      run: dotnet workload install ios
-
-    - name: install workload wasm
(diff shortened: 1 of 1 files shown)


Numpsy:
If it's not one thing it's another :-(
```
ILLINK : error MT0180: This version of Microsoft.iOS requires the iOS 17.0 SDK (shipped with Xcode 15.0).
```

Numpsy:
Is there a reason why the PR builds do ```xcode-select -s /Applications/Xcode_14.3.app/Contents/Developer/``` and the 'publish' builds don't?

JaggerJo:
> Is there a reason why the PR builds do `xcode-select -s /Applications/Xcode_14.3.app/Contents/Developer/` and the 'publish' builds don't?

Hah, good question. I can't recall.

PR #385, opened 2024-01-15
Add static property ClipboardContentBinding to DataGridBoundColumn

GallaFrancesco:
Context from the Avalonia docs: https://reference.avaloniaui.net/api/Avalonia.Controls/DataGridBoundColumn/21520FC9

One issue with DataGrid in FuncUI is that copying is only limited to the whole row. Setting this property should allow users to copy the cells they are selecting.


Numpsy:
The CI failure should hopefully be fixed by picking up #391

JaggerJo:
I'll just merge it as it is. From looking at the changes this should just work.

PR #391, opened 2024-01-30
Try running PR build with xcode 15.1 instead of 14.3

Numpsy:
refs https://github.com/fsprojects/Avalonia.FuncUI/pull/383#issuecomment-1881739616 and the way that the CI run for #385 seems to have broken because of it - so lets see what happens without it there

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -10,9 +10,7 @@ jobs:
     runs-on: macos-13
 
     steps:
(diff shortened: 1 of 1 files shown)


Numpsy:
On a related note, also unclear why the PR builds use macOS-13 and the publish workflow uses macOS-12 ?

Numpsy:
Ok, lets try explicitly using xcode 15.1 instead of 14.3

PR #392, opened 2024-01-30
Update unit testing libraries

Numpsy:
This is pretty trivial, but - 

The NuGet vulnerability reporting feature in the latest Visual Studio reports a warning about a transient dependency in the unit tests project:
image

Updating the unit testing libs to the latest version trims out a huge amount of transitive dependencies on rather old versions of things, which sorts that but also makes the dependencies much cleaner:

image


JaggerJo:
Thanks!

PR #393, opened 2024-01-30
Update Github actions in the PR build pipeline

Numpsy:
I noticed this in the build results when looking at other things:

image

so lets try updating the actions


Numpsy:
CI failed due to #391 I think

Issue #367, opened 2023-11-05
Warnings and errors when publishing a .NET 8 app as NativeAOT


Numpsy:
This is a sort of followup to #281, but with a few FuncUI specific warnings in the build that might have local solutions, so -

I tried updating a FuncUI app to .NET 8 and publishing as NativeAOT with the RC2 SDK.
There are alas loads of warnings from FSharp.Core that are outside our control, but also these ones from FuncUI:

```
4>Avalonia.FuncUI.VirtualDom.Patcher.patchProperty(AvaloniaObject,Delta.PropertyDelta): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperty(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
4>Avalonia.FuncUI.VirtualDom.Patcher.patchProperty(AvaloniaObject,Delta.PropertyDelta): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'System.Reflection.PropertyInfo.PropertyType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
4>Avalonia.FuncUI.VirtualDom.Patcher.create(Delta.ViewDelta): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Activator.CreateInstance(Type,Object[])'. The field 'Avalonia.FuncUI.VirtualDom.Delta.ViewDelta.ViewType@' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
4>Avalonia.FuncUI.VirtualDom.Patcher.create(Delta.ViewDelta): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The field 'Avalonia.FuncUI.VirtualDom.Delta.ViewDelta.ViewType@' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

```
and then this happens at runtime:
```
Unhandled Exception: System.MissingMethodException: No parameterless constructor defined for type 'Avalonia.FuncUI.Component'.
[...]

PR #399, opened 2024-03-08
Add System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers  in a …

Numpsy:
…couple more places

refs #367 

This is some trial and error based on linker warnings and testing rather than a clear 'this was wrong', but on top of the current FuncUI it does appear to be sufficient to get my app running without problems in a NativeAOT build.

Note: For reasons I'm not quite clear on, the compiler wouldn't accept the attributes unless I fully qualified the names, *unless* I remove the ```rec``` from the namespace declaration and put it on the ```Types``` module instead - not sure what's going on there

### DIFF
--- a/src/Avalonia.FuncUI/Types.fs
+++ b/src/Avalonia.FuncUI/Types.fs
@@ -1,12 +1,12 @@
-﻿namespace rec Avalonia.FuncUI
+﻿namespace Avalonia.FuncUI
 
 open Avalonia
 open Avalonia.Controls
 open System
 open System.Threading
-open Avalonia.Data
+open System.Diagnostics.CodeAnalysis
 
-module Types =
+module rec Types =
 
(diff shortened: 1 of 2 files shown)


JaggerJo:
@Numpsy nice seeing some progress on this. Let me know when it's ready to merge.

Numpsy:
Adding the extra attributes plus marking Bogus as a root assembly seems to have got the ContactBook component sample basically working as a .NET 8 Native AOT build - the main app window and the core functionality of the diagnostics tools are both working for me on my Windows test VM (I haven't tried AOT builds on any other platforms).

I have seem one crash when going down a few levels in the diagnostic tools, but that's outside the main library and I think it's hitting another issue I've seen elsewhere with using DataTemplateView in a TreeView, so I'll create a separate discussion about that one.

PR #407, opened 2024-03-24
Change the Elmish music player example to use IStorageProvider instea…

Numpsy:
…d of the obsoleted OpenFileDialog/OpenFolderDialog

As it stands, there are a bunch of build warnings about using the deprecated file / folder picker APIs, and this is an attempt to sort those by changing it to use the newer IStorageProvider interfaces.

Notes:
1) Only changes the picker access, the files/folder are still stored as string file paths
2) IStorageFolder has functions for listing items in the folder, but they're async, so I left ```populateFromDirectory``` mapping back into paths to avoid further changes for now.

### DIFF
--- a/src/Examples/Elmish Examples/Examples.Elmish.MusicPlayer/Songs.fs
+++ b/src/Examples/Elmish Examples/Examples.Elmish.MusicPlayer/Songs.fs
@@ -1,20 +1,19 @@
 namespace Examples.MusicPlayer
 
-
 module Songs =
     open System
     open System.IO
     open Types
+    open Avalonia.Platform.Storage
 
-    let populateSongs (paths: string array): Types.SongRecord array =
+    let populateSongs (paths: IStorageFile seq): Types.SongRecord array =
         paths
-        |> Array.Parallel.map FileInfo
-        |> Array.Parallel.map (fun info -> info.Name, info.FullName)
-        |> Array.Parallel.map (fun (name, path) ->
+        |> Seq.map (fun storageFile ->
             { id = Guid.NewGuid()
-              name = name
-              path = path
+              name = storageFile.Name
+              path = storageFile.Path.LocalPath
               createdAt = DateTime.Now })
+        |> Array.ofSeq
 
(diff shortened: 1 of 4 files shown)


src/Examples/Elmish Examples/Examples.Elmish.MusicPlayer/Songs.fs
```diff
@@ -29,3 +28,9 @@ module Songs =
                   name = name
                   path = path
                   createdAt = DateTime.Now })
+
+    let populateFromFolders(folders: IStorageFolder seq) =
+        folders
+        |> Seq.map _.Path.LocalPath
```
Numpsy, line 34: @JaggerJo are we ok with using F# 8 syntax in the samples, or would you prefer sticking to an older version? (querying after I tried building on a system that only had .NET 7 installed and got build errors)

JaggerJo: I don't mind going with .NET 8.

JaggerJo:
@Numpsy I can test on a modern mac if you want.

Numpsy:
> @Numpsy I can test on a modern mac if you want.

Making it call the open-picker functions through ```Dispatcher.UIThread.InvokeAsync``` has got it working on the old iMac and on Windows and I can try it on Ubuntu later, so if you'd be able to try it on a newer mac that'd be useful (my knowledge of how this interops with ```runWithAvaloniaSyncDispatch``` in Elmish programs is rather thin)

Numpsy:
Anyway, this works for me on

Windows
Ubuntu 22.04 (Gnome / Wayland)
macOS 10.15

When in only worked on Windows before.

@JaggerJo If you're able to test on a current mac before merging that would be nice (I've seen other bugs where the file dialog behaviour got changed in macOS updates, so may as well try even though it's only an example I think)

Numpsy:
After the Avalonia 11.1 update is merged, I'll update this one on top of that and retest

Numpsy:
Updated on top of the .NET 8.0 / Avalonia 11.1 changes.

PR #409, opened 2024-03-27
Remove the hard coded path separators for the music player sample icon

Numpsy:
If I try to run the current version on macOS or Ubuntu it blows up trying to load the icon file, and this seems to sort it


Numpsy:
Will fold this into #407 and try to make that one an 'update and try to make it work better on none-windows platforms' change

PR #410, opened 2024-03-28
Use isNull for null checks

Numpsy:
Some suggestions that I got when testing with the analyzers at https://github.com/ionide/ionide-analyzers/pull/85


JaggerJo:
I actually prefer just having operators. Is there a good reason to use 'isNull' instead?

Numpsy:
It's supposed to generate more efficient code (there's some discussion and links in the FSharpLint issue at https://github.com/fsprojects/FSharpLint/issues/439).

Saying that, quite possibly it doesn't really matter for cases in the diagnostics tools etc

PR #412, opened 2024-03-31
Upgrade to .NET 8

sleepyfran:
CMD + Shift + H: net6 -> net8 :^)


Numpsy:
Not sure about dropping .NET 6.0 support full stop but one specific thing to consider - there are breaking changes in the browser/wasm stuff for .NET 8.0 - the browser projects will need to be .NET 8 to work with Avalonia 11.1 but i'm not sure if the existing Avalona 11.0 code will work on .NET 8 - refs https://github.com/AvaloniaUI/Avalonia/pull/13312

Also I think the CI build would need updating to a new SDK version as well (e.g. https://github.com/fsprojects/Avalonia.FuncUI/blob/4679601afeb52101d584bd4cfc54bea60a3df0db/.github/workflows/pull-requests.yml#L21)

Numpsy:
I was having a go with updating the browser control catalog in https://github.com/Numpsy/Avalonia.FuncUI/tree/users/rw/avalonia_11.1 the other day - maybe needs a call on targetting Avalonia 11.1 in the next version as there are some new properties and controls that could have bindings added as well

JaggerJo:
I don't have any issue with requiring .NET 8. 

@Numpsy do you see issues with that?

Numpsy:
> I don't have any issue with requiring .NET 8.
> 
> @Numpsy do you see issues with that?

Personally no. 

(I've actually been building my Avalonia apps as self contained / single file executables so the version of .NET doesn't really matter at runtime though)

Numpsy:
What's the situation with the web assembly playground example project? (https://github.com/fsprojects/Avalonia.FuncUI/tree/chore/update-to-net8/src/Examples/Component%20Examples/Examples.WebAssemblyPlayground) - I don't see it in the solution, and I guess it'd need to be updated as well?

JaggerJo:
> What's the situation with the web assembly playground example project? (https://github.com/fsprojects/Avalonia.FuncUI/tree/chore/update-to-net8/src/Examples/Component%20Examples/Examples.WebAssemblyPlayground) - I don't see it in the solution, and I guess it'd need to be updated as well?

Can be removed

sleepyfran:
Thanks for the review, folks. I merged with the latest master to integrate @Numpsy's removal of the WASM playground and updated the pipeline's .NET version. I think there shouldn't be anything else on the way, I've tested some of the examples with .NET 8 and I didn't find any issue, but let me know if there's any other concern I can take a look at 😄

sleepyfran:
> Those WASM errors in the build might be because it needs newer versions of SkiaSharp and/or Harfbuzz to work with the newer tools. I did a test build with Avalonia 11.1Beta at #413 that seems to work (Avalonia 11.1 will use the newer WASM libs by default), but I don't know what order these things should best be changed in.
> 
> Update: This change seems enough to get the WASM control catalog building for me locally, without other changes (though not running as that needs changes to the javascript etc as well) - if we want to just get the build going?
> 
> image

Thanks for taking a look! We can actually just move these changes to your PR and have them as part of the upgrade to Avalonia 11.1, I don't think we need to do them separately if that's okay with you 😄 Or would the upgrade still require all the changes you're mentioning to the WASM project?

Numpsy:
> Or would the upgrade still require all the changes you're mentioning to the WASM project?

It appears to just work with the tweaks in the other PR, without needing any other settings changes (I have a FuncUI project at work which I'm trying to update to Avalonia 11.1 to get WASM threading support to work, and that works with the same smaller set of changes as well).

If we're ok with swapping to .NET 8.0 and Avlaonia 11.1 for the next release, then we can either merge both PRs or I can add the changes from here to the other PR and do it all in one

JaggerJo:
👍

sleepyfran:
> If we're ok with swapping to .NET 8.0 and Avlaonia 11.1 for the next release, then we can either merge both PRs or I can add the changes from here to the other PR and do it all in one

Sounds great! I'll abandon this PR then, please just add the changes to your PR and we can merge that one, thanks for the help on the upgrade 😃

Numpsy:
> > If we're ok with swapping to .NET 8.0 and Avlaonia 11.1 for the next release, then we can either merge both PRs or I can add the changes from here to the other PR and do it all in one
> 
> Sounds great! I'll abandon this PR then, please just add the changes to your PR and we can merge that one, thanks for the help on the upgrade 😃

Ok, I've cherry picked your changes into #413

PR #413, opened 2024-04-01
Switch from .NET 6.0 to 8.0, and update to Avalonia 11.1

Numpsy:
Part one of an update to Avalonia 11.1 (whose Browser/WASM support now requires .NET 8.0) - It doesn't add any new bindings 

Also refs #412 as there are breaking changes in the web assembly bits for .NET 8 that require changes on top of just updating the TFM

Updated - rolled in the general update to .NET 8.0 from #412


Numpsy:
@JaggerJo @sleepyfran Do we want to complete this now and then add more updated bindings etc afterwards, or leave it until Avalonia 11.1 is closer to release?

JaggerJo:
👍 I'd complete it

Numpsy:
Ok, will complete it and then update #407 on top and retest that.

Numpsy:
Avalonia 11.1 p2 is out now, so will move up to that version

Numpsy:
Updated to fix merge conflicts, use Avalonia 11.1 beta2, and use the inbox version of SourceLink in the .NET 8 SDK rather than the old v1 nuget packages

Numpsy:
Hmm, seems something isn't getting on with the ```AttachedEventDemo``` sample added in #421

Numpsy:
Updated on top of #422

JaggerJo:
@Numpsy nice! Is this ready to release?

Numpsy:
It all seems to be ok.
I created #431 to track any related extensions or changes there might be.

PR #420, opened 2024-04-19
Add bindings

SilkyFowl:
add bindings for AvaloniaObject ~ TextBox, Flyout

- Add binding of .net event.
- Changed setter and compare function in IAvaloniaList<'T>'s binding to avoid unnecessary updates.
- Code placement is tentative.


src/Avalonia.FuncUI/DSL/Base/IDataTemplateHost.fs
```diff
@@ -0,0 +1,49 @@
...
+                |> dataTemplates.RemoveAll
+
+                for newIndex, template in Seq.indexed templates do
+                    let oldIndex = dataTemplates |> Seq.tryFindIndex template.Equals
+
+                    match oldIndex with
+                    | Some oldIndex when oldIndex = newIndex -> ()
+                    | Some oldIndex -> dataTemplates.Move(oldIndex, newIndex)
+                    | None -> dataTemplates.Insert(newIndex, template)
+
```
SilkyFowl, line 26: I implemented binding for Interfaces here, but would it be better to implement binding either in a normal class or in an abstract class?
It seems that Avalonia's plan is to reduce the number of Interfaces.
If so, I would like to implement binding for the Control class.

src/Avalonia.FuncUI/DSL/Flyout.fs
```diff
@@ -1,31 +1,127 @@
...
+            let factory: AvaloniaObject * ('t * CancelEventArgs -> unit) * CancellationToken -> unit = 
+                (fun (control, func, token) ->
+                    let control = control :?> 't
+                    let hander = EventHandler<_>(fun s e -> func(s :?> 't, e))
+                    let event = control.Closing
+
+                    event.AddHandler(hander)
+                    token.Register(fun () -> event.RemoveHandler(hander)) |> ignore)
+            
+            AttrBuilder<'t>.CreateSubscription(name, factory, func, ?subPatchOptions = subPatchOptions)
```
SilkyFowl, line 123: Should I split it into FlyoutBase.fs/ PopupFlyoutBase.fs?

Numpsy: We have MenuBase in its own file so that would be consistent with menus, but I don't know how important that part really is.

src/Avalonia.FuncUI/DSL/Base/IDataTemplateHost.fs
Numpsy: > If so, I would like to implement binding for the Control class.

As in, make the binding functions against the ```Control``` class instead of the ```IDataTemplateHost``` interface?

src/Avalonia.FuncUI/DSL/Base/StyledElement.fs
```diff
@@ -5,11 +5,165 @@ open Avalonia.Controls.Primitives
...
     open Avalonia.Styling
+    open Avalonia.LogicalTree
+    open System.Threading
+    open System
+
+    module internal ClassesInternals =
+        open System.Linq
+
+        /// pseudoclasse is classe beginning with a ':' character.
+        let isPseudoClass (s: string) = s.StartsWith(":")
```
Numpsy, line 20: s.StartsWith(":") -> s.StartsWith(':') ?

src/Avalonia.FuncUI/DSL/Base/IDataTemplateHost.fs
JaggerJo: I would bind the actual interfaces 👍

src/Avalonia.FuncUI/DSL/Base/StyledElement.fs
```diff
@@ -5,11 +5,165 @@ open Avalonia.Controls.Primitives
...
+
+        /// Update `Classes`'s standard classes with new values. 
+        let patchStandardClasses (classes: Classes) (newValues: string seq) =
```
JaggerJo, line 23: @SilkyFowl could you add a comment here explaining what the problem is any why it makes sense to write a custom differ here? 

(I get it, but would be nice to have it in the sources)

src/Avalonia.FuncUI/DSL/TextBox.fs
```diff
@@ -68,12 +69,75 @@ module TextBox =
...
+        static member onCuttingToClipboard<'t when 't :> TextBox>(func: RoutedEventArgs -> unit, ?subPatchOptions) =
+            AttrBuilder<'t>.CreateSubscription<RoutedEventArgs>(TextBox.CuttingToClipboardEvent, func, ?subPatchOptions = subPatchOptions)
+
+        static member onPastingFromClipboard<'t when 't :> TextBox>(func: RoutedEventArgs -> unit, ?subPatchOptions) =
+            AttrBuilder<'t>.CreateSubscription<RoutedEventArgs>(TextBox.PastingFromClipboardEvent, func, ?subPatchOptions = subPatchOptions)
+
         static member onTextChanged<'t when 't :> TextBox>(func: string -> unit, ?subPatchOptions) =
             AttrBuilder<'t>.CreateSubscription<string>(TextBox.TextProperty, func, ?subPatchOptions = subPatchOptions)
 
+        static member onTextChanged<'t when 't :> TextBox>(func: TextChangedEventArgs -> unit, ?subPatchOptions) =
```
Numpsy, line 135: The second version of onTextChanged seems to have caused some confusion in some of the sample apps, e.g.

image

src/Avalonia.FuncUI/DSL/ItemsControl.fs
```diff
@@ -15,19 +20,64 @@ module ItemsControl =
...
+            let factory: AvaloniaObject * ('t * ContainerPreparedEventArgs -> unit) * CancellationToken -> unit =
+                fun (control, func, token) ->
+                    let control = control :?> 't
+                    let handler = EventHandler<ContainerPreparedEventArgs>(fun s e -> func(s :?> 't, e))
+                    let event = control.ContainerPrepared
+
+                    event.AddHandler(handler)
+                    token.Register(fun _ -> event.RemoveHandler(handler)) |> ignore
+
+            AttrBuilder<'t>.CreateSubscription<'t * ContainerPreparedEventArgs>(name, factory, func, ?subPatchOptions = subPatchOptions)
```
JaggerJo, line 53: Do subscriptions expire here when a new handler is added (say because SubPatchOptions change)?

JaggerJo:
Really good work 👍

There are a few small things we need to do before merging (check comments).

src/Avalonia.FuncUI/DSL/Base/StyledElement.fs
SilkyFowl: I've made the comments more detailed. What do you think? 
Additionally, I'm considering adding a process to verify that pseudoclasses are not included in the binding.

```fs
    /// consider: detail explain comment should be written here?
    module internal ClassesInternals =
        open System.Linq
        
        /// ...

        let validateStandardClasses (classes: string seq) =
            if classes |> Seq.exists isPseudoClass then
                failwith "Pseudoclasses may only setting by the control's protected PseudoClasses property itself."

    type StyledElement with
        /// ...

        /// todo: add comment explaining Pseudoclasses 
        static member classes<'t when 't :> StyledElement>(value: Classes) : IAttr<'t> =
            ClassesInternals.validateStandardClasses value

            let getter: ('t -> Classes) = (fun control -> control.Classes)
            /// ...

        /// todo: add comment explaining Pseudoclasses 
        static member classes<'t when 't :> StyledElement>(classes: string list) : IAttr<'t> =
            ClassesInternals.validateStandardClasses classes
            /// ...
```

src/Avalonia.FuncUI/DSL/Base/StyledElement.fs
SilkyFowl: Exactly, I've made the correction.

src/Avalonia.FuncUI/DSL/Base/StyledElement.fs
JaggerJo: Looks good to me, even tho I was more thinking about explaining why there is special logic for diffing/patching style classes. But we can add that later.

src/Avalonia.FuncUI/DSL/TextBox.fs
SilkyFowl: I would like to exclude the changes to onTextChanged from this PR due to its wide impact. I might submit a separate PR for it later.

src/Avalonia.FuncUI/DSL/ItemsControl.fs
SilkyFowl: Based on the implementation of patchSubscription, it seems that if we register the unsubscription process with CancellationToken, it will be executed properly, but I'm not entirely sure. 

https://github.com/SilkyFowl/Avalonia.FuncUI/blob/7107231775827b79546c5be38deadc4ee8e8ca46/src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs#L21-L49

I'm thinking of adding some test code to verify this.

JaggerJo: Once this is verified we can merge IMO

SilkyFowl:
I've made revisions based on the review comments. Please check

JaggerJo (approved):
LGTM, great work 🎉

SilkyFowl:
Thank you very much.
Please let me know if there is anything I should be doing to merge, because I'm missing out.

JaggerJo:
> Thank you very much. Please let me know if there is anything I should be doing to merge, because I'm missing out.

I just merged it. Assumed the author can merge after a PR is approved.

PR #421, opened 2024-05-03
Add generic RoutedEvent bindings as Attached Event.

SilkyFowl:
There are times when you want to know the current state of the sender in a RoutedEvent.
This PR adds generic bindings that can be used in such cases.


JaggerJo:
Nice work!

Could you provide a sample for this?

SilkyFowl (commented):
Would it be acceptable if the content is a ported version of the information found on [this page](https://docs.avaloniaui.net/docs/concepts/input/routed-events)?

SilkyFowl:
Samples added.
Please check.

JaggerJo (approved):
LGTM

PR #428, opened 2024-06-29
Fix spelling/grammar errors in the docs

matachi:
I was reading the docs and noticed some spelling errors, so I went and fixed those that I saw.


JaggerJo (approved):
LGTM, Thanks for the contribution

Issue #449, opened 2025-02-24
Build errors when compliling with the .NET 9 SDK


Numpsy:
I just tried building the solution with the 9.0.200 .NET SDK (the default/latest on this machine) and got a few build errors about attribute usage. e.g

![Image](https://github.com/user-attachments/assets/3ec97e01-1c63-43f1-9fe5-7ec11c442b2c)

I think this is hitting the issue discussed at https://github.com/dotnet/fsharp/issues/18298#issuecomment-2655973717 where the compiler is enforcing the attributes more than it used to and it needs a newer FSharp.Core to work.

Pinning the build to a .NET 8 SDK with global.json avoids the issue, but otherwise it might need either an FSharp.Core update or attribute changes to fix properly?

JaggerJo:
We should probably bump the version of FShapr. Core then 👍

Numpsy:
Seems to be building ok now with the 9.0.203 SDK as the change that causes it is back to only being enabled if you have LangVersion set to preview, so I don't think any immediate changes are needed here.

PR #450, opened 2025-02-27
fix: remove invalid usages of RequireQualifiedAccess

MadelineRitchie:
This allows building on systems with dotnet 9 installed, and has zero behavioral impact. 

An alternative way to allow building on systems that have dotnet 9 installed would be to generate a global.json file that specifies sdk version 8.


Numpsy:
refs https://github.com/fsprojects/Avalonia.FuncUI/issues/449 - the other possibility is to use FSharp.Core 8.0.300, where RequireQualifiedAccess has had it's Usage attributes fixed to correctly allow it to apply to structs as well as classes

PR #453, opened 2025-03-02
Try to fix compile errors in DragDropDemo

Numpsy:
refs 

image

Should these have been changed in https://github.com/fsprojects/Avalonia.FuncUI/commit/46d44d821093e56da16e114386c4cea02faaf504 ?


JaggerJo:
@Numpsy yes, definitely. 

We should probably have a branch protection policy in place that ensures master always builds.

PR #459, opened 2025-04-21
Add TreeDataGrid bindings

Numpsy:
refs https://github.com/fsprojects/Avalonia.FuncUI/issues/386

Just the bindings for TreeDataGrid, nothing clever.

No docs yet - wanted to get any comments about doing it before spending more time on that.

This puts the bindings in a different library, to avoid a hard dependency on TreeDataGrid for everyone

PR #462, opened 2025-04-27
Move common NuGet package properties into Directory.Build.props

Numpsy:
refs https://github.com/fsprojects/Avalonia.FuncUI/pull/459#discussion_r2052340926

Should PackageVersion be common as well? (I guess the core lib and the elmish lib should be in step, not sure about the diagnostics lib if that gets published again)


JaggerJo:
> refs [#459 (comment)](https://github.com/fsprojects/Avalonia.FuncUI/pull/459#discussion_r2052340926)
> 
> Should PackageVersion be common as well? (I guess the core lib and the elmish lib should be in step, not sure about the diagnostics lib if that gets published again)

IMO it makes sense to keep the version separate.

PR #467, opened 2025-06-14
Update to XUnit v3

Numpsy:
Any thoughts or objections on updating to XUnit v3? (the NuGet package is 'xunit.v3' instead of 'xunit' so it doesn't appear as a direct upgrade option)


src/Avalonia.FuncUI.UnitTests/Program.fs
```diff
@@ -1,6 +0,0 @@
-namespace Avalonia.FuncUI.UnitTests
-
-module Program =
-
-    let [<EntryPoint>] main _ =
```
Numpsy, line 5: XUnit 3 can run tests directly just by doing 'dotnet run' on the project, which generates the entry points for you

image

JaggerJo:
LGTM 👍

PR #476, opened 2025-10-05
[Testing] Avalonia 12

Numpsy:
Avalonia 12 is coming up, so just doing some test builds with it so see if things still build

### DIFF
--- a/src/Avalonia.FuncUI/Avalonia.FuncUI.fsproj
+++ b/src/Avalonia.FuncUI/Avalonia.FuncUI.fsproj
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Include="Avalonia" Version="$(AvaloniaVersion)" />
-    <PackageReference Include="Avalonia.Controls.DataGrid" Version="$(AvaloniaVersion)" />
+    <PackageReference Include="Avalonia.Controls.DataGrid" Version="$(AvaloniaDataGridVersion)" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


JaggerJo:
I’m fine with targeting dotnet 10 👍

Numpsy:
Updated to Avalonia 12 preview 2.

Most of the changes are a result of 

https://github.com/AvaloniaUI/Avalonia/pull/20732
https://github.com/AvaloniaUI/Avalonia/pull/20767

making a number of previously public things internal.

Numpsy:
Updated to Avalonia 12.0.0 release version

Numpsy:
There are a couple of new warnings in the build now

image

I think this is because of a combination of the newer compiler in the .NET 10 SDK being more strict about attribute usage enforcement and the ```RequireQualifiedAccess``` attribute only being marked as applicable to classes in FSharp.Core 6. (it was changed to both classes and structs in FSharp.Core 8)

src/Avalonia.FuncUI/DSL/Buttons/RadioButton.fs
```diff
@@ -13,4 +14,13 @@ module RadioButton =
 
         static member groupName<'t when 't :> RadioButton>(value: string) : IAttr<'t> =
             AttrBuilder<'t>.CreateProperty<string>(RadioButton.GroupNameProperty, value, ValueNone)
-        
\ No newline at end of file
+        
+        static member onChecked<'t when 't :> RadioButton>(func: RoutedEventArgs -> unit, ?subPatchOptions) : IAttr<'t> =
```
Numpsy, line 18: Added this based on comments elsewhere as there are multiple places in the samples that want to know when a radio button is checked but don't care about unchecked events, and it's simpler than duplicating it locally.
Need to decide whether to keep it or not.

Issue #386, opened 2024-01-15
TreeDataGrid Support


GallaFrancesco:
Are there any plans to support [TreeDataGrid](https://github.com/AvaloniaUI/Avalonia.Controls.TreeDataGrid) in FuncUI?

It should be possible to use it in place of `DataGrid` with the additional benefits of having `TreeView` on top and other perks such as `SelectableTextBlock` by default on `DataGridTextColumn`.

Numpsy:
I've been making some simple use of it in a FuncUI application and there could be some 'official' bindings for it, though I don't think the core library should depend on it.

GallaFrancesco:
Why not? The `DataGrid` control is supported and it's managed by the Avalonia organization as well.

JaggerJo:
Would you contribute the bindings?

I'm not entirely happy with how we included the bindings for data grid in the main library. This creates a transitive dependency on Avalonia.Controls.DataGrid for all projects referencing FuncUI.

We should have small binding packages for control libraries.

Numpsy:
TreeDataGrid also has a dependency on System.Reactive, which is quite large itself, so I don't think forcing a dependency is ideal (not my descision of course)

JaggerJo:
> TreeDataGrid also has a dependency on System.Reactive, which is quite large itself, so I don't think forcing a dependency is ideal (not my descision of course)

💯 agreed.

Numpsy:
Hmm, had a look at my usage of TreeDataGrid and it's basically all code behind style and no real bindings - e.g. doing things like
```fsharp
let makeTreeDataSource initialItems =
    let gridSauce =
        new HierarchicalTreeDataGridSource<StructuredSectionContent>(items = initialItems)

    gridSauce.Columns.Add(
        HierarchicalExpanderColumn<StructuredSectionContent>(
            TextColumn<StructuredSectionContent, string>("Name", (fun section -> section.DisplayName)),
            (fun x -> x.SubSections),
            (fun x -> not (Seq.isEmpty x.SubSections))
        )
    )

    gridSauce.Columns.Add(
        TextColumn<StructuredSectionContent, string>("Value", (fun section -> section.DisplayValue))
    )

    gridSauce
```
and then just assigning that to the ```Source``` property of the grid in an ```init``` function.

Numpsy:
[...]

PR #500, opened 2026-04-15
[Test] Split the DataGrid bindiungs out into their own project.

Numpsy:
## Description

Just testing for now, I know there are issues with calling a new NuGet package 'Avalonia.*'
(Base idea being to make the base library not depend on the seperate Avalonia.Controls.DataGrid library, so only people who actually want it can pull it in)

## Related Issue

refs https://github.com/fsprojects/Avalonia.FuncUI/issues/386


JaggerJo:
we could move the packages to AvaloniaCommunity

Numpsy:
could do

Numpsy:
> we could move the packages to AvaloniaCommunity

Are you able to create a new repo over there? Would need to do it before a v12 based release to avoid more breaking changes later

JaggerJo:
> > we could move the packages to AvaloniaCommunity
> 
> Are you able to create a new repo over there? Would need to do it before a v12 based release to avoid more breaking changes later

You should have received an e-mail invite

Numpsy:
https://github.com/AvaloniaCommunity/FuncUI-Community-Bindings/pull/1

PR #511, opened 2026-05-08
Use XCode 26.5 for CI builds

Numpsy:
refs the CI failre for #510 

image

Looks like the build agent updating to a newer .NET SDK means that it requires a newer XCode version.

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -11,11 +11,11 @@ concurrency:
 jobs:
   build:
 
-    runs-on: macos-15
+    runs-on: macos-26
 
     steps:
     - name: xCode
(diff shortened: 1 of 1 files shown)


Numpsy:
Except it seems the macOS 15 build agent doesn't have 26.4 on it.
Not sire if thats just because it hasn't been updated or if the build needs to use macOS 26

Saying that, the build for the master branch is working and that's on macOS 15 as well

Numpsy:
According to https://github.com/actions/runner-images/issues/13977, Xcode 26.4 requires macOS 26 which might require an update if the latest .NET iOS tools require Xcode 26.4

Numpsy:
This appears to be a recurring thing now - 
image

26.5 will apparently be made the default in a week https://github.com/actions/runner-images/issues/14172

JaggerJo (approved):

PR #514, opened 2026-07-11
Start adding some bindings for the new TableView control

Numpsy:
## Description

This is a new control added in Avalonia 12.1.
Also added a super simple page for it in the control catalogue

refs https://github.com/AvaloniaUI/Avalonia/issues/21237 / https://github.com/AvaloniaUI/Avalonia/pull/21511

## Checklist

- [x] I have tested my changes locally
- [x] My changes build without errors
- [x] All existing tests pass

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <AvaloniaVersion>12.0.0</AvaloniaVersion>
+    <AvaloniaVersion>12.1.0</AvaloniaVersion>
     <FuncUIVersion>2.0.0-preview1</FuncUIVersion>
   </PropertyGroup>
 	
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - name: xCode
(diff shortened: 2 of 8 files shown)


Numpsy:
We could do something a bit more advanced than this, but it does show the most basic example working

image

src/Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <AvaloniaVersion>12.0.0</AvaloniaVersion>
+    <AvaloniaVersion>12.1.0</AvaloniaVersion>
```
Numpsy, line 3: This does require Avalonia 12.1 instead of 12.0

JaggerJo:
Nice 👍

Numpsy:
The CI seems to have failed because .NET 10 is wanting another newer version of XCode again :-(

image

(According to https://github.com/actions/runner-images/issues/14344 the default will change from 26.5 to 26.6 next week)

Numpsy:
@JaggerJo Are you ok with the 2.0 version going straight to Avalonia 12.1, or would you prefer to stick with 12.0 to start with and do an update afterwards?

JaggerJo:
I'm fine with that 👍

JaggerJo (approved):

PR #515, opened 2026-07-25
Set version to 2.0.0-preview2

Numpsy:
Just wondering whether updating from Avalonia 12.0 to 12.1 makes it worth doing another preview build?


JaggerJo:
I'd be fine with another preview or releasing directly. Your call :)

JaggerJo (approved):

Numpsy:
I'll complete this, then give it a test in the work project I wanted to add the Table bindings for :-)

Issue #427, opened 2024-06-21
App.axaml replacement


CoolCoderSuper:
Currently you can't define styles or resources in app using the DSL which isn't that big of a deal since it is pretty easy to add manually.
However it is more annoying when I want to have a TrayIcon.
```xaml
<TrayIcon.Icons>
	<TrayIcons>
		<TrayIcon Icon="Assets/icon.ico">
			<TrayIcon.Menu>
				<NativeMenu>
					<NativeMenuItem Header="Settings"/>
					<NativeMenuItem Header="Start"/>
					<NativeMenuItem Header="Stop"/>
					<NativeMenuItem Header="Import"/>
					<NativeMenuItem Header="Exit"/>
				</NativeMenu>
			</TrayIcon.Menu>
		</TrayIcon>
	</TrayIcons>
</TrayIcon.Icons>
```

JaggerJo:
Are there no bindings for TrayIcon yet?

If not we should add them. Are you willing to contribute the bindings?

CoolCoderSuper:
I would be willing to try.
We would also need to add bindings for NativeMenu as well.
https://github.com/fsprojects/Avalonia.FuncUI/blob/master/docs/controls/nativemenu.md

I'm not sure how one would go about registering it on the window or application since the DSL creates IView<T> which wouldn't work with `NativeMenu.SetMenu`

CoolCoderSuper:
If I understand correctly the Component type is a real Avalonia component that renders out the IViews.
However neither NativeMenu or TrayIcon are actual Avalonia components.

JaggerJo:
@CoolCoderSuper should still work as the [base type is Avalonia Object](https://github.com/AvaloniaUI/Avalonia/blob/3b235f9aa82dbcae024c0e75486e7d91df1be13d/src/Avalonia.Controls/NativeMenu.cs#L10).

`IView` is probably not the best name as all types inheriting from Avalonia Object are supported.

JaggerJo:
>I'm not sure how one would go about registering it on the window or application since the DSL creates IView which wouldn't work with NativeMenu.SetMenu or TrayIcon.SetIcon

If it works in XAML it should also work in the DSL without explicitly calling those methods.

CoolCoderSuper:
OK that makes sense.
The xaml compiler generates calls to those methods for it to work.

JaggerJo:
Hah, ok. We could add 'register' calls that internally use init functions...

[...]

PR #483, opened 2025-11-15
Update to Avalonia v11.3.9 and Implement ResourceDictionary DSL

SilkyFowl:
This PR updates Avalonia from v11.1.0 to v11.3.9. With this version update, `ResourceDictionary` now inherits from `AvaloniaObject`, enabling us to implement ResourceDictionary-related DSL bindings.

### Avalonia Version Update
- Updated Avalonia from v11.1.0 to v11.3.9 (Directory.Build.props)

### ResourceDictionary DSL Implementation
New file: ResourceDictionary.fs

- **`ResourceDictionary.keyValue`**: Supports setting key-value pairs
  - Provides overloads for both object values (`obj`) and IView values
  - Implements clear functionality using `ClearValue` type
- **`ResourceDictionary.mergedDictionaries`**: Supports merged dictionaries
- **`ResourceDictionary.themeDictionariesKeyValue`**: Supports theme variant-specific resource dictionaries

### ResourceInclude Wrapper Implementation
Since `ResourceInclude` does not inherit from `AvaloniaObject`, we implemented a `ResourceIncludeWrapper` class:
- Wraps `ResourceInclude` with a class that inherits from `ResourceProvider`
- Recreates the internal instance when the `Source` property changes to ensure resource updates
- Provides helper functions: `fromUri`, `fromString`

### ResourceProvider DSL Implementation
New file: ResourceProvider.fs

- **`onResourceObservable`**: Binding that subscribes to `GetResourceObservable`
  - Supports optional `defaultThemeVariant` parameter
  - Wraps results in `option` type (treating `null` and `UnsetValue` as `None`)

### StyledElement DSL Extensions
StyledElement.fs

- **`StyledElement.resources`**: Added overload that accepts `IView`
- **`StyledElement.onResourceObservable`**: Resource observation binding for StyledElement

### Test Infrastructure Improvements
- **Introduced `Avalonia.Headless.XUnit` package**: Required for testing `ResourceInclude`
- **Added ResourceDictionaryTests.fs**: Comprehensive test suite (493 lines)
- **Added `TestHelpers.fs`**: Assertion helpers for tests
- **Updated existing tests**: Significantly expanded resource-related tests in `StyledElementTests.fs`
- **Added test resource files**: `TestResources1.xaml`, `TestResources2.xaml`

### Minor Fixes
- `AttrBuilder.fs`: Fixed typos in documentation comments
[...]

PR #517, opened 2026-09-10
Add stateful application hosting and native menu/tray bindings

CoolCoderSuper:
## Summary

This is a continuation of PR #473, towards fixing #427.

Add a stateful, XAML-free way to configure the application itself, including its native menu and tray icons.

`HostApplication` patches the existing application instance created by `AppBuilder`. `ApplicationComponent` reuses FuncUI's state and effect hooks through a nonvisual `IStateContext`. It does not create a hidden control or replace `Application.Current`.

- Add application bindings for name, requested theme, native menu and tray icons.
- Expand native menu/item bindings with submenus, separators, icons, gestures, toggle type, visibility, tooltips and lifecycle/click events. Add tray icon and click bindings.
- Support subscriptions on `AvaloniaObject`, removing the previous `Control` restriction.
- Cancel native subscriptions on removal and replacement, and cooperate with Avalonia's attached tray collection to dispose each platform handle once.
- Dispose application hooks on controlled-lifetime exit and skip queued renders/effects after disposal.
- Give `ComponentBase` the same pending-render coalescing as the application host.

## Small example

```fsharp
let application =
    ApplicationComponent(fun ctx ->
        let count = ctx.useState 0
        let menu = NativeMenu.create [
            NativeMenu.items [
                NativeMenuItem.create [
                    NativeMenuItem.header $"Increment ({count.Current})"
                    NativeMenuItem.onClick (fun _ -> count.Set(count.Current + 1))
                ]
            ]
        ]
        [
            Application.nativeMenu menu
            Application.trayIcons [
                TrayIcon.create [
                    TrayIcon.icon ()
                    TrayIcon.toolTipText $"Count: {count.Current}"
                    TrayIcon.menu menu
                ]
            ]
        ])

type App() =
    inherit HostApplication()

    override this.Initialize() =
        this.Styles.Add(FluentTheme())

    override this.OnFrameworkInitializationCompleted() =
        match this.ApplicationLifetime with
        | :? IClassicDesktopStyleApplicationLifetime as desktop ->
            let window = new HostWindow()
            window.Content <- ()
            desktop.MainWindow <- window
            desktop.ShutdownMode <- ShutdownMode.OnMainWindowClose
            this.Update(application)
        | _ -> ()
        // Registers automatic component/native-object cleanup on lifetime exit.
        base.OnFrameworkInitializationCompleted()
```

## Breaking changes and migration

- **Public subscription type changes.** `Types.Subscription.Subscribe` now takes `AvaloniaObject * Delegate` rather than `Control * Delegate`. Custom subscription implementations with explicit `Control` parameters must accept `AvaloniaObject` and cast when they specifically require a control. Rebuild dependent assemblies.
- **Virtual DOM update signature changes.** `VirtualDom.update` now accepts `AvaloniaObject` instead of `Control`. Ordinary calls passing controls remain source-compatible, but the CLR signature changes, so dependent assemblies must be rebuilt.
- **Context interface changes.** State/effect members move from `IComponentContext` to its new base interface, `IStateContext`. Normal component hook calls remain source-compatible. Custom context implementations and reflection-based integrations may need updates; F# implementations should implement the inherited state members under `IStateContext`.
- **Render scheduling changes.** `ComponentBase` coalesces requests made while a render is pending. `AfterRender` effects run per actual render, not per request, so code relying on redundant renders or effect counts may behave differently. A request made during rendering still schedules another pass.
- **Disposed work no longer runs.** Detached components skip queued renders, and disposed effect queues discard pending effects. Do not rely on effects or forced renders after component disposal.
- **Native content removal now performs cleanup.** Removed/replaced native content loses its FuncUI subscriptions, and removed tray icons release their platform handles. Treat objects created by native content bindings as host-owned; do not retain removed tray icons for reuse. Supplied icon images remain caller-owned.

## Validation

- All 39 unit tests pass, including native subscription replacement/removal, state retention, tray disposal, nested cleanup and component scheduling.
- The new example, existing CounterApp and Elmish library build successfully.

Run the sample:

```sh
dotnet run --project src/Examples/Examples.Application
```

## Next Steps

While I think this already feels way better than before, it still doesn't completely eliminate XAMLesque style.

It doesn't have a way to register styles or the framework init.

I kind of feel like in order to the rest of it well, we would need a proper Window and Style dsl.


src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs
```diff
@@ -13,12 +13,36 @@ module internal rec Patcher =
...
+        let subscriptions = ViewMetaData.GetViewSubscriptions target
+        if subscriptions <> null then
+            for subscription in subscriptions.Values do
+                subscription.Cancel()
+                subscription.Dispose()
+            subscriptions.Clear()
+
+    // Native objects have no logical-tree detach notification. Release only objects
+    // owned by native content bindings, not arbitrary user-supplied IDisposable values.
+    let internal releaseNative (value: obj) : unit =
```
JaggerJo: Why exactly do we need special logic for TrayIcon, NativeMenu, ...?

src/Avalonia.FuncUI/VirtualDom/VirtualDom.Patcher.fs
```diff
@@ -86,8 +112,26 @@ module internal rec Patcher =
...
-                collection.Clear()
+                match collection with
+                | :? TrayIcons ->
+                    // RemoveAt supplies OldItems even when Clear would emit Reset.
+                    while collection.Count > 0 do
+                        releaseItem collection.[collection.Count - 1]
+                        collection.RemoveAt(collection.Count - 1)
+                | _ ->
+                    for item in collection do releaseItem item
+                    collection.Clear()
```
JaggerJo: same here, not a fan of having type specific code in the patcher

src/Examples/Examples.Application/App.fs
```diff
@@ -0,0 +1,132 @@
...
+
+    override this.Initialize() =
+        this.Styles.Add(FluentTheme())
+        // Register custom styles once, after the theme, rather than on each render.
+        this.Styles.Add(
+            StyleInclude(
+                Uri("avares://Examples.Application/"),
+                Source = Uri("avares://Examples.Application/AppStyles.axaml")))
+
+    override this.OnFrameworkInitializationCompleted() =
```
JaggerJo: bootstrapping now seems a lot more complicated. Can we simplify this? 

lIf we want to merge this the code required should become simpler overall - also for the other sample apps.

JaggerJo:
@Numpsy what do you think?

Numpsy:
I'll try to make some time over the weekend to have a look, have some things going on at the moment.

> It doesn't have a way to register styles or the framework init.

I like the idea of being able to do everything in F# rather than needing xaml, but I do sometimes put control style references in app.axaml because it works better with self contained executables and trimming (if the only reference to a style is a string in the code then the trimmer might remove it, but if it's referenced from a compiled app.axaml then it gets kept), so it'd be nice if there was a way to improve that as well.

CoolCoderSuper:
That makes sense.
The way I did it one sample if by putting the styles in a separate xaml file then adding it the init function of the App.

```fsharp
this.Styles.Add(
            StyleInclude(
                Uri("avares://Examples.Application/"),
                Source = Uri("avares://Examples.Application/AppStyles.axaml")))
```

Numpsy:
> That makes sense. The way I did it one sample if by putting the styles in a separate xaml file then adding it the init function of the App.
> 
> ```fsharp
> this.Styles.Add(
>             StyleInclude(
>                 Uri("avares://Examples.Application/"),
>                 Source = Uri("avares://Examples.Application/AppStyles.axaml")))
> ```

Maybe something like https://github.com/Numpsy/Avalonia.FuncUI/commit/253dd1b5e65476cdedd7ef0f61fd49da2bb20cf2 would work - If I build the current example as a single file with trimming enabled it crashes on startup due to missing compiled xaml, but doing that instead seems to get it working.

src/Avalonia.FuncUI/DSL/AttrBuilder.fs
```diff
@@ -287,9 +287,9 @@ type AttrBuilder<'view>() =
 
         // subscribe to avalonia property
         // TODO: extract to helpers module
-        let subscribeFunc (control: Control, _handler: 'h) =
+        let subscribeFunc (control: AvaloniaObject, _handler: 'h) =
             let cts = new CancellationTokenSource()
-            control
+            (control :?> Control)
```
Numpsy, line 292: Do we need to consider things that aren't controls getting into here?

Numpsy:
> It doesn't have a way to register styles or the framework init.
> I kind of feel like in order to the rest of it well, we would need a proper Window and Style dsl.

Actually, looks like some of the subscription changes here are the same as was proposed in https://github.com/fsprojects/Avalonia.FuncUI/pull/483, which was doing things with resources?

src/Avalonia.FuncUI/DSL/AttrBuilder.fs
CoolCoderSuper: The GetObservable(routedEvent) is defined as an extension method on the Interactive type.
I could make it cast to Interactive instead of control and that would be more accurate unless we would create a new implementation of that extension of the AvaloniaObject.

[src/Avalonia.FuncUI/Library.fs:64](https://github.com/fsprojects/Avalonia.FuncUI/blob/master/src/Avalonia.FuncUI/Library.fs#L64)

CoolCoderSuper:
> Maybe something like https://github.com/Numpsy/Avalonia.FuncUI/commit/253dd1b5e65476cdedd7ef0f61fd49da2bb20cf2 would work - If I build the current example as a single file with trimming enabled it crashes on startup due to missing compiled xaml, but doing that instead seems to get it working.

I did not test this with trimming, thanks.
I will do some testing with that and see what I find.

> Actually, looks like some of the subscription changes here are the same as was proposed in https://github.com/fsprojects/Avalonia.FuncUI/pull/483, which was doing things with resources?

Interesting I may use it as a reference if I ever try doing a resources DSL.

